PR #6, opened 2016-07-14
Add table block to specification

nlisgo:
@thewilkybarkid

### DIFF
--- a/src/blocks/table.v1.yaml
+++ b/src/blocks/table.v1.yaml
@@ -1,0 +1,14 @@
+$schema: http://json-schema.org/draft-04/schema#
+title: Table
+type: object
+properties:
+    type:
+        type: string
+        enum:
+          - table
+    html:
+        title: HTML table
+        type: string
+required:
+  - type
+  - html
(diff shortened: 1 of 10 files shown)


thewilkybarkid:
Looks good; you'll need to execute `node compile.js` and commit the results.

src/blocks/table.v1.yaml
```diff
@@ -0,0 +1,14 @@
+$schema: http://json-schema.org/draft-04/schema#
+title: Table
+type: object
+properties:
+    type:
+        type: string
+        enum:
+          - table
+    html:
+        title: HTML
```
thewilkybarkid: Might be worth expanding on this. 'HTML table'?

thewilkybarkid:
@nlisgo Following #8 could you `npm install` then `node compile.js`. This should remove `dist/scripts/api-console-vendor.js` from being included.

nlisgo:
@thewilkybarkid I rebased and confirm that my PR no longer includes changes to `dist/scripts/api-console-vendor.js`

PR #12, opened 2016-07-25
Don't try and use name parts

thewilkybarkid:
As we know, [names are hard](https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/). Recently we've had a case of an author without a (possibly-known) given name.

So, I think we should avoid trying to use named parts of names (as that's near impossible to get right), and instead use a few basic forms. I propose 'long' (note this is different to 'full') and 'short', and also include an 'index' form for sorting.

### DIFF
--- a/dist/model/person-list.v1.json
+++ b/dist/model/person-list.v1.json
@@ -25,25 +25,27 @@
                                 "title": "Name",
                                 "type": "object",
                                 "properties": {
-                                    "givenNames": {
-                                        "title": "Given names",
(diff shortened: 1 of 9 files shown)


thewilkybarkid:
Spoken to @lsh-0 about this and we think it's simpler to just have 'preferred' rather than 'long' and 'short'. I don't think there's a need for the latter at the moment, though we can always introduce it later.

PR #15, opened 2016-08-01
Add decision letters and author responses

thewilkybarkid:


src/model/article-vor.v1.yaml
```diff
@@ -23,10 +23,44 @@ allOf:
...
+                    minItems: 1
+                content:
+                    type: array
+                    items:
+                        $ref: ../misc/blocks-all.v1.yaml
+                    minItems: 1
+            required:
+              - doi
+              - description
+              - content
```
thewilkybarkid, line 44: `decisionLetter` and `authorResponse` should be next to each other. Had to reorder due to https://github.com/BigstickCarpet/json-schema-ref-parser/issues/24.

PR #20, opened 2016-08-01
Add article status everywhere

thewilkybarkid:
@lsh-0 has requested having the article `status` property included in the full article version.

We had to include this in listings to distinguish between PoA and VoR already, but individually the content type would have had to be used.

### DIFF
--- a/src/model/search.v1.yaml
+++ b/src/model/search.v1.yaml
@@ -10,87 +10,42 @@ properties:
         title: Search results
         type: array
         items:
-            allOf:
-              - type: object
-                required:
-                  - type
-              - oneOf:
-                  - allOf:
-                      - properties:
-                            type:
-                                type: string
-                                enum:
-                                  - correction
-                                  - editorial
-                                  - feature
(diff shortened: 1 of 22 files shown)


src/model/search.v1.yaml
```diff
@@ -10,87 +10,42 @@ properties:
...
-                  - allOf:
-                      - properties:
-                            type:
-                                type: string
-                                enum:
-                                  - podcast-episode
-                      - $ref: ../snippets/podcast-episode.v1.yaml
+            oneOf:
+              - $ref: ../snippets/article-vor.v1.yaml
+              - $ref: ../snippets/article-poa.v1.yaml
```
thewilkybarkid, line 15: I don't know why these have to be in this order, but the JSON references don't appear to be resolved properly otherwise. Must be an upstream bug somewhere...

thewilkybarkid, line 15: Opened https://github.com/BigstickCarpet/json-schema-ref-parser/issues/24.

PR #33, opened 2016-08-15
Update description for index name

nlisgo:


thewilkybarkid:
The preferred/index values are there to allow for the name structure to appear as the person would want it to be in an English-language publication (see #12). In this case that example is actually wrong; closing in favour of #35.

PR #38, opened 2016-08-24
Turn image into complete figures

thewilkybarkid:
This should be all of the extra data needed for figures.

The only thing I'm aware of is the different sizes for display.

Ping @gnott.


gnott:
The fig image support looks good and something to build toward in the parser.

PR #47, opened 2016-09-09
Give examples of BC breaks

thewilkybarkid:


src/api.raml
```diff
@@ -20,17 +20,43 @@ documentation:
     content: |
         API endpoints are not versioned, but the body of requests/responses are.
 
-        Content versioning
-        ------------------
+        ## Content versioning
 
         Endpoints produce/consume dedicated content type, with the version as a parameter (eg `application/vnd.elife.labs-experiment-list+json;version=1`). These are primarily JSON-based (and described using [JSON Schema](http://json-schema.org/)), but alternative representations may be available.
 
         This verisoning follows [semantic versioning](http://semver.org/). Note we only use the major version as minor versions are backwards compatabile (so the API will return versions `1` and `2` rather than `1.0`, `1.1`, `2.0` etc).
```
giorgiosironi, line 27: typo here

src/api.raml
```diff
@@ -20,17 +20,43 @@ documentation:
...
+          - Removing a property
+          - Adding a new possibility to an array property
+
+        ### On a response
+
+        The following examples *are* breaking changes (and so *require* a new version):
+
+          - Making a required property optional
+          - Adding a new possibility to an array property
+          - Removing a possibility from an array property
```
giorgiosironi: Not sure about this last one

thewilkybarkid: Yeah, that's not right (would be true on a request).

PR #49, opened 2016-09-13
File uri pattern

gnott:
Should the file uri also start with https://  ?

### DIFF
--- a/src/misc/file.v1.yaml
+++ b/src/misc/file.v1.yaml
@@ -26,6 +26,7 @@ properties:
         title: Source data URI
         type: string
         format: uri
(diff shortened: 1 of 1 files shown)


thewilkybarkid:
I only included a HTTPS-requirement on those that are expected to be embedded (eg images)... but maybe we should do it everywhere for consistency.

gnott:
This change wasn't required to pass validation, and I only thought of it when rewriting the values to be valid uri. I just wanted to ask. I'm happy to close this and not merge it if it is not required to be https:// (even though we will use https:// in eLife cases).

PR #50, opened 2016-09-15
Standardise eLife IDs

thewilkybarkid:
Limits all content IDs (that aren't natural ones) to lowercase letters and numbers.

### DIFF
--- a/src/misc/id.v1.yaml
+++ b/src/misc/id.v1.yaml
@@ -1,0 +1,4 @@
+$schema: http://json-schema.org/draft-04/schema#
+title: eLife ID
+type: string
+pattern: ^[a-z0-9-]+$
--- a/src/snippets/article.v1.yaml
+++ b/src/snippets/article.v1.yaml
@@ -3,8 +3,7 @@ title: Article snippet
 type: object
 properties:
     id:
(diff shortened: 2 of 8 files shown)


src/samples/article-poa/v1/complete.json
```diff
@@ -16,8 +16,8 @@
     },
     "pdf": "https://elifesciences.org/content/5/e14107.pdf",
     "subjects": [
-        "biochemistry",
-        "biophysics-structural-biology"
+        "bioch",
```
giorgiosironi: why hyphens (and hence multiple separable words) are forbidden?

thewilkybarkid: Mainly to avoid delimiting problems if we use them in URIs and use a hyphen to separate it from a slug. Though we could always use a slash instead.

(The original subject IDs here I just made up based on the name, not sure if that's a good idea in case they get renamed, though I'm not sure if we would ever rename them. Plus we probably wouldn't use a subject ID + slug combination...)

thewilkybarkid: We've agreed offline to allow for hyphens.

PR #58, opened 2016-09-21
Add IDs to articles sections

thewilkybarkid:
This adds an optional ID to `section` blocks, but makes it required for immediate article sections.


gnott:
I'll just note here for information: some eLife article body is not wrapped in a sec tag. For now, I am adding a section block that doesn't exist. I will also add an id to that phantom section. Just an FYI.

PR #59, opened 2016-09-21
Change podcast episode MP3 to sources

thewilkybarkid:
Not that we can at the moment, this would allow us to have multiple formats for podcast episodes (eg OGG).

/cc @nlisgo

### DIFF
--- a/src/samples/collection/v1/complete.json
+++ b/src/samples/collection/v1/complete.json
@@ -149,7 +149,7 @@
             },
             "sources": [
                 {
-                    "mediaType": "audio/mp3",
+                    "mediaType": "audio/mpeg",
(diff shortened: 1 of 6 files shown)


giorgiosironi:
Interestingly, this seems to be failing after merge:
https://ci--alfred.elifesciences.org/job/library-api-raml/43/console

thewilkybarkid:
Think it needed a rebase. I've just required that PRs are up to date to avoid this in the future.

giorgiosironi:
The version of Jenkins that we have now is correctly setting the commit status with the `continuous-integration/jenkins/pr-head` name which was a prerequisite, so you can turn that on for all projects that you need.

PR #60, opened 2016-09-22
ELPP-1483 List improvements

thewilkybarkid:
This allows the different prefix characters, and for any block type (including multiple blocks). This should now match the requirements of articles.

Ping @gnott and @nlisgo.


gnott:
Looks good. I can change the ordered -> prefix logic once you've merged, and see if the new block elements are validating now.

PR #62, opened 2016-09-22
First test for setting commit status manually

giorgiosironi:


giorgiosironi:
What I need has not been executed here, but I needed the PR to get this into `develop` due to the status check. I may need to turn on and off the check for debugging.

PR #68, opened 2016-09-27
Triggering a rebuild of dist/ on Alfred

giorgiosironi:
This rebuild was necessary due to previous rebuilds not committing removed files in dist/


giorgiosironi:
Merging to trigger the new behavior.

PR #69, opened 2016-09-27
ELPP-1521 Add subject snippets

thewilkybarkid:
Thoughts @gnott, @nlisgo, @stephenwf?


gnott:
Should be ok on the article XML parsing side. I believe it is currently converting the string value to a slug-like value (Luke did this already before I started on it), so we just need to also include the full text value as well and change the data structure. :+1:

stephenwf:
Looks good, hopefully this will remove the need for extra network calls when normalising too:

https://github.com/elifesciences/api-sdk-php/blob/master/src/ApiSdk.php#L107-L109

PR #79, opened 2016-10-18
ELPP-1630 Add 'unknown' references

thewilkybarkid:


gnott:
Can we add more specific optional fields to an 'unknown' reference in the future?

PR #80, opened 2016-10-18
added 'versionDate' and 'stage' attributes to article model

lsh-0:


lsh-0:
> "Date/time the `status` changed to this value (eg article moved from PoA to VoR).\n"

hrm, I rewrote the description, but the original is actually more nuanced. the status date in the case where no change in status has occurred is the same as the pub date? so a v2 POA statusDate is the same as the v1 POA until a VOR v3 is published?

PR #83, opened 2016-10-26
Proposal: stick to one format for date-time time-offset

giorgiosironi:
[https://www.ietf.org/rfc/rfc3339.txt|RFC 3339] allows two formats for
the time-offset of a UTC date-time:
- `Z`
- `+00:00`
  I could not find a functional difference between the two.

However, it would cause less confusion down the road if we stick to one
of these formats. PHP's `datetime` extension using `DATE_ATOM` as a
format produces `+00:00`, so I would choose that.

Immediate advantage: PHP code in tests can generate exactly these
JSON samples, rather than dumping slightly different strings for
`date-time` fields.

The diff is just a sample, but it would be easy enough to complete it
with `s/00Z/00+00:00/g` on `src/samples/`.

### DIFF
--- a/src/samples/collection/v1/complete.json
+++ b/src/samples/collection/v1/complete.json
@@ -74,8 +74,8 @@
             "doi": "10.7554/eLife.09560",
             "authorLine": "Lee R Berger et al",
             "title": "<i>Homo naledi</i>, a new species of the genus <i>Homo</i> from the Dinaledi Chamber, South Africa",
-            "published": "2015-09-10T00:00:00Z",
-            "statusDate": "2015-09-10T00:00:00Z",
+            "published": "2015-09-10T00:00:00+00:00",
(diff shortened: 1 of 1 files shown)


thewilkybarkid:
Probably a good idea. I'd like to enforce UTC-everywhere too. So, we should not use `format: date-time` directly, and instead reuse a separate schema like we have for DOIs, ISBNs etc (this _would_ use `format: date-time`, as well as a `pattern` to check that it ends with `+00:00`).

giorgiosironi:
Ok, I'll look into that and update.

lsh-0:
Python `datetime.isoformat` produces `+00:00` out of the box as well, which is what Lax is using  until I mustered an effort to find a rfc3339 formatter

giorgiosironi:
We are now using `Z` as the only suffix.

PR #88, opened 2016-11-08
Add press-packages

nlisgo:


nlisgo:
@thewilkybarkid I thought I would start this off with a pull request which borrows heavily from the blog articles. It is my understanding that we can reference at least one article, in fact we may require at least one article. Also, am I right in thinking that the subjects should be gathered from the referenced articles rather than input on the press package level?

thewilkybarkid:
Looking good. Almost certain that they have to reference at least 1 article. And the subjects of the press package are computed from the articles (not sure about ordering, maybe article 1 subjects + article 2 subjects + ... then make them unique?), but for the API definition that doesn't matter (the implementation can always be changed to be curated list without the schema changing).

nlisgo:
I've added support for relatedContent. The media contacts could use the `misc/person.v1.yaml` but add the email and telephone property?

src/model/press-package.v1.yaml
```diff
@@ -0,0 +1,28 @@
...
+                oneOf:
+                  - $ref: ../snippets/article-poa.v1.yaml
+                  - $ref: ../snippets/article-vor.v1.yaml
+            minItems: 1
+            uniqueItems: true
+        mediaContacts:
+            title: Media contacts
+            type: array
+            items:
+                $ref: ../misc/person.v1.yaml
```
nlisgo: @thewilkybarkid We need to allow an array of people with the option properties of affiliation, email addresses and phone numbers. Please can you advise on the best approach for this.

thewilkybarkid: We should do something like `author.v1.yaml` and add in the extra properties (so like https://github.com/elifesciences/api-raml/blob/develop/src/misc/author.v1.yaml#L31-L35 and https://github.com/elifesciences/api-raml/blob/develop/src/misc/author.v1.yaml#L46-L49). So here it'd be an `allOf` with the first item being the `$ref` and the second item being the two new properties.

thewilkybarkid: So in the style of https://github.com/elifesciences/api-raml/blob/develop/src/misc/author.v1.yaml#L55-L61

nlisgo: Cool. Will give this a go.

src/model/press-package.v1.yaml
```diff
@@ -0,0 +1,43 @@
...
+                  - $ref: ../snippets/article-vor.v1.yaml
+            minItems: 1
+            uniqueItems: true
+        mediaContacts:
+            title: Media contacts
+            type: array
+            items:
+                allOf:
+                  - $ref: ../misc/person.v1.yaml
+                  - properties:
```
nlisgo, line 28: @thewilkybarkid how does this look? I will extend the sample fixtures once you are ok with this.

src/model/press-package.v1.yaml
```diff
@@ -0,0 +1,43 @@
...
+                  phoneNumbers:
+                      type: array
+                      items:
+                          $ref: phone-number.v1.yaml
+                  postalAddresses:
+                      type: array
+                      items:
+                          $ref: address.v1.yaml
+    required:
+      - content
```
thewilkybarkid, line 43: I suggest we make `relatedContent` required. If we later find out it's optional we can always change it.

src/model/press-package.v1.yaml
```diff
@@ -0,0 +1,43 @@
...
+            items:
+                allOf:
+                  - $ref: ../misc/person.v1.yaml
+                  - properties:
+                      emailAddresses:
+                          type: array
+                          items:
+                              type: string
+                              format: email
+                  phoneNumbers:
```
thewilkybarkid: Indentation needs fixing.

src/model/press-package.v1.yaml
thewilkybarkid: 👍

nlisgo:
@thewilkybarkid I believe I have addressed your latest feedback.

nlisgo:
Just going to rebase

src/snippets/press-package.v1.yaml
```diff
@@ -0,0 +1,28 @@
+$schema: http://json-schema.org/draft-04/schema#
+title: Press package snippet
+type: object
+properties:
+    id:
+        description: ID
+        type: string
+        minLength: 1
```
thewilkybarkid: I've just merged #50, could you change this to a `$ref: ../misc/id.v1.yaml`?

nlisgo: Sure

nlisgo:
@thewilkybarkid I have extended the sample json fixture. I believe that affiliations is more appropriate than postalAddresses. Please review.

PR #97, opened 2016-11-15
Article history must contain complete article version snippets

nlisgo:


thewilkybarkid:
Looks good. I think we should add a `description` on `versions` to clarify the ordering.

src/api.raml
```diff
@@ -280,7 +280,7 @@ traits:
                 Individual article history.
             get:
                 description: |
-                    Get an article history.
+                    Get an article history, sorted by version number in descending order.
```
nlisgo: @thewilkybarkid, Is this description adequate?

thewilkybarkid: Sorry, I meant in `article-history.v1.yaml` (there isn't a description yet).

src/model/article-history.v1.yaml
```diff
@@ -9,25 +9,12 @@ properties:
         type: string
         format: date
     versions:
+        description: List of articles, sorted by version in descending order
         type: array
```
nlisgo, line 13: @thewilkybarkid How is this description?

thewilkybarkid: It's a list of 'versions' not 'articles'.

thewilkybarkid:
@lsh-0 are you happy with this change?

src/samples/article-history/v1/unpublished-multiple-versions.json
```diff
@@ -2,8 +2,71 @@
...
+        {
+            "status": "vor",
+            "id": "14107",
+            "version": 3,
+            "type": "research-article",
+            "doi": "10.7554/eLife.14107",
+            "authorLine": "Yongjian Huang et al",
+            "title": "Molecular basis for multimerization in the activation of the epidermal growth factor",
+            "stage": "preview",
+            "published": "2016-03-28T00:00:00Z",
```
thewilkybarkid, line 14: ~~Need to lose this.~~

lsh-0:
so - you're replacing what is there with the snippet?

src/samples/article-history/v1/unpublished-multiple-versions.json
thewilkybarkid: Sorry, this is fine!

nlisgo:
@lsh-0 that is what I'm hoping for. If we can get this in then journal-cms (and perhaps search) does not need to retain a knowledge of what values should be present in an article snippet. We will just store the data snippet which is available in the /article/{id}/versions response.

lsh-0:
ok. it probably won't make it into lax until end of week

thewilkybarkid:
Happy for this to be merged when Lax is ready.

lsh-0:
it should be ok to merge now because lax and the bot-lax-adaptor use pinned revisions of the api-raml to test against. 

of course - if the journal site is validating incoming article-json against latest version of the raml we will have problems. pinging @giorgiosironi

PR #99, opened 2016-11-16
ELPP-1767 Refactor figures

thewilkybarkid:
(This would replace #98.)

We have a new requirement to allow videos as figure supplements. As @davidcmoulton is making the figure viewer flexible in what it can display, I think it would be useful to make the representation here flexible. /cc @gnott @nlisgo @lsh-0

@Melissa37 and @JGilbert-eLife, it would be useful to know if conceptually the following is how you consider figures/tables/videos (bearing in mind we're trying to share the same way of doing things between articles and non-articles).

I'm calling images, tables and videos (and audio files in the future) an 'asset'. These can be captioned. Captioning can be as little as adding a title, but it can extend to include a HTML ID, a label, text beneath the title and attribution. This covers inline images.

These assets can then be placed into a 'figure'. The first asset is the primary asset, any others are supplementary. These assets must be fully captioned (except for text beneath the title and an attribution, which are optional). They must also have a DOI, and may have source data (which are captioned like assets, but are downloadable files). This is what the figure viewer will cover, and what appear on the article figures page.

This means that figures are now quite flexible: a video figure _could_ have a table supplement, a table figure _could_ have an image supplement, an image figure _could_ have a video supplement. (I don't think we actually have any non-image figures with supplements, but as soon as we need one we don't have to do anything.)

In some non-article content people are trying to include figure-style images (eg https://elifesciences.org/elife-news/inside-elife-what-it-costs-publish). We're not currently planning to include the fancy figure viewer anywhere but on articles, so I don't think they should have access to add it (technically this schema will allow it). They can have access to fully caption an asset, which is probably good enough. (I don't think I've seen anyone try an add figure supplements/source data... yet.)


lsh-0:
is there a jira ticket and is this part of the article deliverables?

thewilkybarkid:
ELPP-1767.

thewilkybarkid:
For clarity, this adds 'figure' as a separate block type, rather than the figure viewer using the presence of a label etc to trigger.

Melissa37:
I think this sounds OK to me. Thank you.
M

gnott:
I think this covers the fig, figsupp, fig figsupp data, fig video, combinations I'm aware of. It also potentially covers more combinations as you say.

It would be helpful to coordinate this change if it will be merged soon. I'd like some time to develop the XML to JSON conversion on a separate branch and get it most of the way done before the RAML schema changes.

thewilkybarkid:
@gnott I've made some tweaks so that both `title` and `caption` are now optional, so we can just use what's in the XML. Changed `label` slightly so that it's only part of `figure` assets (ie having a label forces use of the `figure` type). /cc @nlisgo

gnott:
Q. Should figures be brought into a "figure" block after this change, instead of being an "image" block?
Q. Is the provision for sourceData now removed from tables and videos? I can find 1 video with source data, and more than 100 tables with sourceData.

thewilkybarkid:
Inline images/video/tables remain as their current type, ones that are 'figures' (ie have a `label`) need to be moved into a `figure` block.

`sourceData` now applies to `figure` assets only (don't think we ever include it on inline ones?).

gnott:
I have an early draft of changes to the parser that will wrap assets in a "figure" as outlined here.

One issue so far, testing with the 00666 kitchen sink, the table with id="inlinetable1" has no label, and this schema requires table to have a label (I believe, at least that is what the validation error is saying).

Is this change to the RAML schema planned to be merged soon? I see there are currently some merge conflicts with this branch.

I would still need to update all my test fixtures to reflect the code changes I'm testing with first.

thewilkybarkid:
@gnott It only requires an `id` if it's becoming a `figure` type not a `table` type.

thewilkybarkid:
https://github.com/thewilkybarkid/api-raml/blob/1a974992d382f406f3e9e18f4cdc793a964ce9ce/src/samples/article-vor/v1/00569.json#L1435-L1441 is a table that has an `id` but not a `label`.

gnott:
I'll see if I can add the logic to understand the inline table versus a figure table and wrap it appropriately. The sample XML is a good illustration.

giorgiosironi:
Is this backward compatible? Will this need to be merged before bot-lax-adaptor or journal starts using the new schema or later?

thewilkybarkid:
It is (I think). If Journal starts using it now, all figures appear like inline images etc (since they're not the `figure` type yet).

gnott:
We could test backwards compatibility by rolling back the elifetools version used in https://github.com/elifesciences/bot-lax-adaptor/pull/204  and all should be valid, if we want to test it out.

thewilkybarkid:
🎉

PR #110, opened 2016-11-24
Sections may not have empty titles

giorgiosironi:
Journal asserts this:
```
[2016-11-24 11:52:05] request.CRITICAL: Uncaught PHP Exception
Assert\InvalidArgumentException: "Value "" is blank, but was expected to
contain a value." at
/srv/journal/vendor/beberlei/assert/lib/Assert/Assertion.php line 220
{"exception":"[object] (Assert\\InvalidArgumentException(code: 27):
Value \"\" is blank, but was expected to contain a value. at
/srv/journal/vendor/beberlei/assert/lib/Assert/Assertion.php:220)
[stacktrace]
Assert\\Assertion::createException('', 'Value \"\" is bla...', 27, NULL)
/srv/journal/vendor/elife/patterns/src/ViewModel/ArticleSection.php(38):
Assert\\Assertion::notBlank('')
/srv/journal/vendor/elife/patterns/src/ViewModel/ArticleSection.php(82):
eLife\\Patterns\\ViewModel\\ArticleSection->__construct('phantom-s-12',
NULL, '', 2, '<p><b>Related r...', false, true, false)
eLife\\Patterns\\ViewModel\\ArticleSection::collapsible('phantom-s-12',
'', 2, '<p><b>Related r...', false, false)

```
so we should check it at this level?

### DIFF
--- a/src/blocks/section.v1.yaml
+++ b/src/blocks/section.v1.yaml
@@ -11,6 +11,7 @@ properties:
     title:
         title: Text
         type: string
+        minLength: 1
     content:
         description: Content
         type: array


giorgiosironi:
This was the article triggering the problem, being generated from `elife-15853-vor-r1`:
```
elife@end2end--journal:~$ curl -v http://end2end--gateway.elifesciences.org/articles/1578514472 | python -m json.tool
* Hostname was NOT found in DNS cache
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying 54.91.137.33...
* Connected to end2end--gateway.elifesciences.org (54.91.137.33) port 80 (#0)
> GET /articles/1578514472 HTTP/1.1
> User-Agent: curl/7.35.0
> Host: end2end--gateway.elifesciences.org
> Accept: */*
>
< HTTP/1.1 200 OK
< Date: Thu, 24 Nov 2016 11:54:30 GMT
< Content-Type: application/vnd.elife.article-vor+json;version=1
< Transfer-Encoding: chunked
< Connection: keep-alive
* Server nginx/1.4.6 (Ubuntu) is not blacklisted
< Server: nginx/1.4.6 (Ubuntu)
< Vary: Accept
< X-Frame-Options: SAMEORIGIN
< Allow: OPTIONS, GET
< KONG-Authenticated: False
< X-Kong-Upstream-Latency: 43
< X-Kong-Proxy-Latency: 0
< Via: kong/0.8.3
<
{ [data not shown]
100 15847    0 15847    0     0   202k      0 --:--:-- --:--:-- --:--:--  203k
* Connection #0 to host end2end--gateway.elifesciences.org left intact
{
    ...
    "body": [
        {
            "content": [
                ...
                {
                    "text": "The question still remains as to which aspects of neural communication we are missing by not being able to routinely record neural activity directly from the surface of the human brain. So we are surely in for more surprises, whether we like it or not.",
                    "type": "paragraph"
                }
            ],
            "id": "phantom-s-12",
            "title": "",
            "type": "section"
        }
```

thewilkybarkid:
:+1:. There's probably other cases like this.

giorgiosironi:
@lsh-0 @gnott this means a change is needed in bot-lax-adaptor or elife-tools, in order to always fill in this field with some content.

gnott:
There's a placeholder title right now added only if title is missing. I can make it so a placeholder is added if the value is a blank string as well.

It's odd you have a "phantom-s-12" id, as the counter should start at "phantom-s-1" for each article generated. I wonder if how the script is called the global counter is not getting set back to 1? That might cause some randomness across the article versions with the ID values changing all the time. Those are only added on sections that have no ID.

giorgiosironi:
Loading this from the testing machine with `curl -v https://end2end--lax.elifesciences.org/api/v2/articles/1578514472 | python -m json.tool` I can see there are no other `phantom-...` ids in the JSON, so I too think is a global state issue.

PR #112, opened 2016-11-25
Validating date-time values

giorgiosironi:
`misc/date-time.v1.yaml` is used for all fields that were relying on the `date-time` standard. We find the official format too loose as we want to impose that all dates are UTC, they have no milliseconds/microseconds part, and the UTC timezone is indicated with `Z`.


src/misc/date-time.v1.yaml
```diff
@@ -0,0 +1,9 @@
+$schema: http://json-schema.org/draft-04/schema#
+title: Date-time ISO8601 with numeric UTC timezone
+description: |
+    eLife restriction over the ISO8601 standard
+
+    For example:
+        - 2016-09-16T12:34:56Z
+type: string
+pattern: ^[1-2][0-9]{3}(-((0[1-9])|(1[0-2]))(-((0[1-9])|([1-2][0-9])|(3[0-1]))))T((2[0-3])|([01][0-9])):[0-5][0-9]:[0-5][0-9]Z$
```
thewilkybarkid: We should be able to combine `format` and `pattern`, having the latter than just checks the length.

PR #113, opened 2016-11-28
CR-51 Make funding awards optional

thewilkybarkid:


gnott:
Looks good I think thanks Chris!

PR #114, opened 2016-12-01
Forces images URLs to end with a good extension

giorgiosironi:
I got all the ones that were already forced to be starting with `https://`

### DIFF
--- a/src/blocks/image.v1.yaml
+++ b/src/blocks/image.v1.yaml
@@ -46,7 +46,7 @@ definitions:
                 title: Image URI
                 type: string
                 format: uri
-                pattern: ^https://
+                pattern: ^https://.*\.(tif|jpg)
             attribution:
                 title: Attribution
                 type: array
--- a/src/blocks/image.v1.yaml
+++ b/src/blocks/image.v1.yaml
@@ -46,7 +46,7 @@ definitions:
                 title: Image URI
                 type: string
                 format: uri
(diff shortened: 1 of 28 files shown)


giorgiosironi:
@davidcmoulton are there other extensions that will be valid for images coming from the API?

gnott:
I think we'll need .gif support at a minimum too for animated gifs. Seems restrictive to only allow certain types although it should support most of eLife article cases I can remember.

giorgiosironi:
Will investigate how to extract this pattern into a single place in `src/`. Also we may want to do a similar thing for podcast and video URLs.

thewilkybarkid:
Can we park this for a week or two? We need to change how we handle images anyhow to cater for various sizes. (For display purposes we definitely don't want a TIFF, but the downloadable source could be anything.)

giorgiosironi:
Sort of, but the bug needs to be fixed in the bot-lax-adaptor or an unspecified percentage of articles will be missing images. This is just a way to validate it, doing it at the `end2end` level is limited to a sample of 15 articles.

thewilkybarkid:
Hmm, not entirely convinced by this (might be missing some context). I think this should just care that it's a (HTTPS) URI, whether it has a file extension or not doesn't really matter (e.g. if they weren't being served straight from a file system we might not have them). Sounds like the real problem is that the URI doesn't resolve.

PR #117, opened 2016-12-12
Fix ISBN regex

thewilkybarkid:
We only want hyphenated ISBN-13s. /cc @gnott 

Based on https://www.safaribooksonline.com/library/view/regular-expressions-cookbook/9781449327453/ch04s13.html.


gnott:
Thanks Chris. I think all the current ISBN are passing validation, after the isbn 10 value I removed the hyphen. There are no specific rewriting rules in the parser for ISBN outside of that one example, so it is probably most important that @Melissa37 is aware of the allowed formats for production. If there are unacceptable ISBN values then it is probably best that they do raise a validation exception.

Melissa37:
@thewilkybarkid , I think we'll need to write a rule in the Schematron, but I still don't really understand what the rule is! Can you explain?

Thanks!

lsh-0:
Closing due to lack of activity. Feel free to re-open if ISBNs are still an issue.

PR #121, opened 2017-01-03
ELPP-1972 Allow any block type in acknowledgements

thewilkybarkid:


gnott:
Will this be ok to merge soon? I believe this is why 12741 v2 JSON is not valid.

PR #122, opened 2017-01-04
ELPP-1912 Change all images to IIIF

thewilkybarkid:
This replaces all images with a IIIF Image API endpoint (and access to the original source file).

I've include the sample file from https://rucore.libraries.rutgers.edu/api/?api=iiif everywhere, which is a IIIF Image API v2 server (we should use our own when available).

This doesn't yet include focal point information, which will be important for banners/thumbnails.


src/blocks/image.v1.yaml
```diff
@@ -21,14 +21,8 @@ definitions:
...
           - properties:
-                alt:
-                    title: Alternative text
-                    type: string
-                uri:
-                    title: Image URI
-                    type: string
-                    format: uri
-                    pattern: ^https://
+                image:
```
giorgiosironi, line 24: This is nuclear for the bot-lax-adaptor, could we prepare an intermediate version where:
- `alt` and `uri` are allowed but optional
- `image` is allowed but optional
- if it's easy: at least one of the two options (`alt`+`uri` or `image`) is required
That would give the bot and lax complex time to migrate to the new format without breaking all of `prod--journal.elifesciences.org` images. It won't be immediate to implement DLCS integrations that are able to produce this `image` field.

thewilkybarkid: This is a proposal based on a discussion yesterday, coordination is needed on deploying and we shouldn't do that yet.

We spoke about generating static files in the CDN in the IIIF URI syntax for the sizes that we require, which would allow us to generate all the images ahead of time, switch over, then later down the line use a proper IIIF server.

thewilkybarkid:
I've added the properties required for focal points. I've added the height/width of the image to avoid having to read every image `info.json`.

thewilkybarkid:
I've added a commit that replaces `video` block images with a IIIF one (and renames to `placeholder`). This assumes that we can get the image from Glencoe/Exeter.

thewilkybarkid:
Note I'd expect the image and video sizes to be the same in real life.

giorgiosironi:
Who is discovering how we can get those images? If they are in the article .zip package, they should be served with the rest. But if they are on Glencoe, some additional configuration is needed in iiif (or some additional activity in the bot to retrieve them).

thewilkybarkid:
@Melissa37 is currently clarifying with Exeter. A bot activity to fetch them is best, but in theory pointing Loris and Glencoe would work (relies on Glencoe not changing things).

gnott:
I suggest the bot to grab them and put copies onto S3 https://elifesciences.atlassian.net/browse/ELPP-2373

PR #123, opened 2017-01-12
Fix data sets

thewilkybarkid:
This will cause some articles to start to fail (eg 17082). /cc @gnott @lsh-0


giorgiosironi:
After we merge, I'll open a PR in bot-lax-adaptor to raise the number of allowed invalid articles using this new version. We expect about ~15 more, all problems come from the `date` field, sometimes it's `'NA'`, sometimes `null`, sometimes it's missing.

giorgiosironi (approved):

PR #126, opened 2017-01-25
ELPP-2161 Add seperate citation metrics

thewilkybarkid:
We don't what the citations metrics to be time-based, so this separates it out to what we actually need.

/cc @lsh-0


lsh-0:
cheers, this is looking more manageable

PR #127, opened 2017-01-27
ELPP-2032 Allow basic tables in captions

thewilkybarkid:
I've disallowed some properties to force the table to a minimum.

This also then won't trigger a figure viewer inside a figure viewer (#99 would solve this, but these tables should probably be caption-less anyway).

/cc @gnott


gnott:
Thanks for this Chris. It looks like it will solve the one case and make it clean for future use :+1:

PR #128, opened 2017-01-30
ELPP-1850 Add figures PDF

thewilkybarkid:
cc @gnott @lsh-0

### DIFF
--- a/src/samples/article-list/v1/first-page.json
+++ b/src/samples/article-list/v1/first-page.json
@@ -28,6 +28,7 @@
             "volume": 4,
             "elocationId": "e09560",
             "pdf": "https://elifesciences.org/content/4/e09560.pdf",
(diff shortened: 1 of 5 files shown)


gnott:
I think this might be somewhere in lax and not in the bot-lax-adaptor, so probably over to @lsh-0 for more guidance of the next steps.

lsh-0:
um - so without much context, is this about the CDN? if so, then yes - ticket is assigned to me. I'll take a look at what is left to do today

giorgiosironi:
Should still be in bot-lax-adaptor as it's part of the JSON (fragments) being generated there?

gnott:
I guess to be more accurate I meant the data is not in the XML, so it doesn't apply to the "first level" of scraping which only looks at the XML (which is the only one I'm directly involved in right now).

PR #129, opened 2017-01-31
ELPP-1323 Add highlights API

thewilkybarkid:
This doesn't define what lists are available: `magazine`, `community` + subject IDs are valid (the latter's dynamic, so can't be defined here).

(Note the specification for subject highlights are that they're generated rather than manually curated.)

We have a slight problem that we need to include podcast episode _chapters_ (ie part of a bit of content). I've tried to make it generic by having a `fragment` property. This would normally be a HTML ID (eg to an article figure), but for podcast episode chapters the fragment is their `time` property.

I'm not entirely happy with that approach, so might look at alternatives.

/cc @nlisgo


thewilkybarkid:
Thinking about it, there's a change to recommendations that we need to make so that has chapters rather than episodes as well, so we should do a proper solution.

thewilkybarkid:
The `podcast-episode-chapter` ends up quite big, maybe we should remove the `content`. (So this is a 'chapter snippet'.)

src/samples/highlights/v1/complete.json
```diff
@@ -0,0 +1,213 @@
...
+                        "name": "Ecology"
+                    }
+                ]
+            },
+            "chapter": {
+                "number": 2,
+                "title": "Lost generation",
+                "time": 300,
+                "impactStatement": "Is science at risk of losing talented researchers to other professions?",
+                "content": [
```
giorgiosironi: I agree that also for consistency with the other types we should not include `content` but rather a `podcast-episode-snippet`; this is how content types are always included when they are referred in list.

It isn't how podcast episode chapters are listed in podcast full models, but all these API definitions follow patterns of usage so if `content` and other fields are not necessary they may be dropped here.

PR #130, opened 2017-02-01
Simplify tables in captions

thewilkybarkid:
We're currently stuck on a version of the PHP JSON Schema library that gets stuck in an infinite loop (causing https://github.com/elifesciences/api-sdk-php/pull/98 and https://github.com/elifesciences/api-sdk-php/pull/99 to fail). Newer versions of the library handle it correctly, but while we wait for https://github.com/webmozart/json/pull/27 to be merged this simplifies the schema.

Use of `additionalProperties` is problematic, so we haven't been using it, but for this case it's probably ok (especially temporarily, and hopefully can be reverted before 1.0 is locked).

### DIFF
--- a/src/misc/asset.v1.yaml
+++ b/src/misc/asset.v1.yaml
@@ -21,15 +21,19 @@ properties:
             oneOf:
               - $ref: ../blocks/mathml.v1.yaml
               - $ref: ../blocks/paragraph.v1.yaml
-              - allOf:
-                  - $ref: ../blocks/table.v1.yaml
-                  - properties:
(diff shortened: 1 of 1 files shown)


gnott:
I've tried this as a test and it fails on elife-05375-v2.xml, the article with a table in the caption. I think it is because the table has an id attribute, and this schema does not allow anything but "type" and "tables". Is is possible to also allow "id"? Or, should I remove the id when parsing it? It seems strange for a caption table to have an id, but maybe it is important.

PR #131, opened 2017-02-02
Allow IDs on tables in captions

thewilkybarkid:
#130 reduced the amount of properties allowed in captions in tables, including IDs that are included in the XML. Normally I'd say lose it as nothing references it, and I don't know if we'd be able to support it properly when presenting, but we already allow quite a bit of MathML that have IDs too, and I've no idea if anything's referencing those.

/cc @gnott

### DIFF
--- a/src/misc/asset.v1.yaml
+++ b/src/misc/asset.v1.yaml
@@ -27,6 +27,8 @@ properties:
                         type: string
                         enum:
                           - table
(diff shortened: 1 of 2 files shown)


gnott:
Tested and it fixed the issue, all valid now - thanks @thewilkybarkid!

PR #132, opened 2017-02-02
ELPP-2095 Add date ranges

thewilkybarkid:
This add date ranges to `/covers` and `/search`. I've not include times, as we don't need that granularity (we're still talking UTC though).

Having a `endDate` before a `startDate`, or having either in the future, needs to result in a `400 Bad Request` (alongside invalid formats).

Ping @nlisgo, @stephenwf.

### DIFF
--- a/src/api.raml
+++ b/src/api.raml
@@ -149,13 +149,13 @@ traits:
                 example: cell-biology
     dateRanged:
         queryParameters:
-            startDate:
+            start-date:
                 description: |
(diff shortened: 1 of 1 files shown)


stephenwf:
Just a quick question about invalid dates, should we throw a `422` or a more generic `400` ?

thewilkybarkid:
Follow RFC 7231 and make it a `400`. (A `422` refers to the body content anyhow.)

PR #134, opened 2017-02-06
ELPP-2246 Move related articles to a separate API

thewilkybarkid:
I haven't made it a paginated list as articles tend to have a couple at most.


giorgiosironi:
The separate resource should make serving these data from Lax a lot simpler. Not sure how bot-lax-adaptor will communicate this to Lax, but there's probably going to be a specialized command that it uses to "upsert" these relations during ingestion.

giorgiosironi (approved):
Not sure if this can be merged in a backward-compatible way, however.

PR #140, opened 2017-02-10
ELPP-2241 Add author type to book references

thewilkybarkid:
Other way of doing this is a boolean, but this matched clinical trials.

/cc @gnott


gnott:
Would this mean for a book, if there are authors and editors, then we can only include the authors? Maybe having authors and editors (such as in book-chapter) were never supported, but I believe I'm adding both types anyway.

PR #142, opened 2017-02-14
ELPP-2284 Add published/updated dates to all normal content types

thewilkybarkid:
For consistency, this adds `published` and `updated` to all normal content types (except articles... though I wonder if `versionDate` should become `updated`).

We need to (usually) order collections by `updated`, but this does change it so `updated` is optional. If it's not present (ie never been updated), `published` would be used. We _could_ have it so `updated` is required, in which case it would be the same as `published`. To tell if something is updated you would have to see if they're the same, rather than whether `updated` is present.

/cc @nlisgo @stephenwf @giorgiosironi

### DIFF
--- a/src/samples/blog-article-list/v1/first-page.json
+++ b/src/samples/blog-article-list/v1/first-page.json
@@ -12,6 +12,7 @@
             "title": "Media coverage: Slime can see",
             "impactStatement": "In their research paper – Cyanobacteria use micro-optics to sense light direction – Schuergers et al. reveal how bacterial cells act as the equivalent of a microscopic eyeball or the world’s oldest and smallest camera eye, allowing them to ‘see’.",
             "published": "2016-07-08T08:33:25Z",
+            "updated": "2016-07-09T08:33:25Z",
             "subjects": [
                 {
                     "id": "biophysics-structural-biology",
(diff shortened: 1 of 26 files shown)


giorgiosironi:
I agree with having a consistent field that is always required for ordering, it makes easier to define queries and indexes on that field rather than on a combination. In fact, IIRC search creates ordering fields in itself. The combination of the two fields being initialized with the same value is pretty common, e.g. https://github.com/KnpLabs/DoctrineBehaviors/blob/master/src/Model/Timestampable/TimestampableMethods.php

That said, I wouldn't make it `required` in this PR so that applications have time to adapt, but later down the line.

For articles, the field could be added in the same way, deprecating the older one `versionDate`, adding the description of `updated`; changing lax and bot-lax-adaptor; requiring `updated` while forbidding `versionDate`.

thewilkybarkid:
I'm tempted to leave articles for at least the moment. They are a bit different in that they have versions, so `versionDate` as a name is explicit (and `updated` sounds like it would include silent corrections). Not sure adding a _fourth_ date to articles is a sensible idea, but we might have to review if with ever add a `use-by=updated` (would that be latest `versionDate` or `statusDate`?).

PR #144, opened 2017-02-20
ELPP-2304 Allow book references to have both authors and editors

thewilkybarkid:
@gnott Please merge when you're happy/ready.


gnott:
Looks good I think and we can merge now. I'm just tweaking the output to take advantage of it now.

PR #145, opened 2017-02-21
ELPP-2309 Simplify journal names

thewilkybarkid:
Rather than attempting to include publisher details, this just makes it the name.

/cc @lsh-0 @gnott Please merge if/when happy.


lsh-0:
ah, thanks for this. I have no problems implementing it, I just needed an example to work with.

PR #146, opened 2017-02-22
Make periodical a simple string

thewilkybarkid:
Following on from #145, the same should apply to periodical names.

/cc @gnott


gnott:
Thanks @thewilkybarkid ready to go!

PR #147, opened 2017-03-07
ELPP-2393 Add external events

thewilkybarkid:
/cc @nlisgo

### DIFF
--- a/src/api.raml
+++ b/src/api.raml
@@ -696,6 +696,9 @@ traits:
                                 complete:
                                     displayName: Complete
                                     value: !include samples/event/v1/complete.json
+                                complete-external:
+                                    displayName: Complete external event
+                                    value: !include samples/event/v1/complete-external.json
 
 /highlights/{list}:
     type: collection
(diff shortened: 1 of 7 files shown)


thewilkybarkid:
Could get rid of `venue` completely (at least for the moment).

thewilkybarkid:
I've removed `venue`. I've also then simplified the `Place` schema to remove Google Place ID/coordinates as we don't use it anywhere now (/cc @gnott this is used in a few article places, but shouldn't affect anything as they don't have that data).

thewilkybarkid:
Unrelated, but since I'm thinking about `place`, I'm wondering about making `components` optional. I know there are some places in articles that we are only sourcing it from a string, so don't have the granular data. /cc @gnott

gnott:
A couple places I can find when searching these changes would affect
1. Funding source - I think we only populate the "name" value of a place, so that should be ok here
2. Author affiliations -> address -> components specifies the country. The country name is included in address->formatted but it is just a string, although that is probably not an issue for viewing by a person - it may be less machine readable though if you wanted to extract the country specifically

nlisgo:
I'm happy with these changes. I welcome the removal of venue as it significantly reduces the complexity of the UI to maintain events.

PR #153, opened 2017-04-07
Require an ID if there's a DOI

thewilkybarkid:
We're going to point DOIs as fragments, so we'll always need an ID.


thewilkybarkid:
@gnott I don't think this should break anything, happy to merge?

gnott:
It's hard to tell since we have many ``<boxed-text>`` with no id, but we are usually unwrapping it. I guess the best way to tell is merge this, and we can to a full generation test to find the stragglers.

PR #157, opened 2017-04-26
Adding sample of captionless image

giorgiosironi:
See also https://github.com/elifesciences/journal/pull/409


NuclearRedeye:
@lsh-0 update and merge or close?

lsh-0:
api-raml changes should be strictly documented, they affect many applications.

I don't know what prompted this change but it would be good to pin down it's origin before closing.

lsh-0:
Might be related to this change by @nlisgo : https://github.com/elifesciences/patterns-php/blob/503ff586e2abf7f7524d881adaf64c82f4e2ff3a/tests/src/ViewModel/CaptionedAssetTest.php

NuclearRedeye:
@nlisgo Can you shed any light on this PR?

nlisgo:
> @nlisgo Can you shed any light on this PR?

This can be merged in. No schema change here. This just verifies that an image with a caption is valid.

lsh-0:
@nlisgo , if this is still good to merge and doesn't interfere with the social images work, could you merge it please? I think it should be safe.

nlisgo:
@lsh-0 can you confirm you are happy for this to be a minor version bump?

NuclearRedeye:
> @lsh-0 can you confirm you are happy for this to be a minor version bump?

@nlisgo As this seems to only be adding an example rather that changing the API do we need to bother bumping the version number? I'm not the maintainer, but my 2 cents would be that the version only needs to apply to the API and hence in this case it should be fine to leave it as is.

lsh-0:
yeah, another example of not needing a version bump.

PR #162, opened 2017-05-15
Extend person snippet

nlisgo:


thewilkybarkid (changes requested):
This means that anything using the snippet has all the new properties, like collections. Instead, we should change the person list to use the model rather than the snippet. That should then be a 1-liner here and in the SDK.

nlisgo:
@thewilkybarkid good point. I'll refactor.

thewilkybarkid (approved):

PR #168, opened 2017-05-25
ELPP-2741 Add person affiliations

thewilkybarkid:


nlisgo:
This looks good. I just want to confirm that the [address](https://github.com/elifesciences/api-raml/blob/develop/src/misc/address.v1.yaml) can just be a country because that's the only information that we have other than a place name.

thewilkybarkid:
Yeah, we only have that elsewhere too.

PR #175, opened 2017-07-03
ELPP-2756 Related content for press packages is now optional

nlisgo:


src/api.raml
```diff
@@ -1126,15 +1126,15 @@ traits:
                     description: |
                         Press package.
                     body:
-                        application/vnd.elife.press-package+json;version=1:
-                            schema: !include ../dist/model/press-package.v1.json
+                        application/vnd.elife.press-package+json;version=2:
```
thewilkybarkid: We need to leave in the version 1 as we'll support it for a period of time (as least the duration of the switchover).

src/model/press-package.v2.yaml
```diff
@@ -0,0 +1,49 @@
...
+                $ref: ../misc/blocks-all.v1.yaml
+            minItems: 1
+        relatedContent:
+            title: Related content
+            type: array
+            items:
+                oneOf:
+                  - $ref: ../snippets/article-poa.v1.yaml
+                  - $ref: ../snippets/article-vor.v1.yaml
+            minItems: 1
```
thewilkybarkid: This needs to go.

thewilkybarkid (approved):
Looks good.

Do we know if a press pack would ever need to link to a non-article? Since the original spec has changed, we should find out now to try and avoid a new version in the future.

PR #176, opened 2017-07-05
ELPP-2542 Move attribution to a higher level

thewilkybarkid:
We need to have attribution on all images. Currently it's part of the `image` block (through `asset`); this is an attempt to move it to image itself, so that it can be reused. (The slight change is that the attribution for an image block moves from the `attribution` property to `image.attribution`, but it was optional anyway.)


giorgiosironi:
PR is clear enough, `asset` is broken into `attribution` and `caption` so that blocks can use the two independently. A bit inconsistent to have the `image` block different from the others but it helps the reuse of the `image` misc.

giorgiosironi:
Preparing for merging.

giorgiosironi:
(but https://github.com/elifesciences/bot-lax-adaptor/pull/271 should go first)

PR #178, opened 2017-07-17
ELPP-1000 Job adverts API

giorgiosironi:


src/model/job-advert.v1.yaml
```diff
@@ -0,0 +1,45 @@
...
+allOf:
+  - $ref: ../snippets/job-advert.v1.yaml
+  - properties:
+        description:
+            description: Description
+            type: array
+            items:
+                $ref: ../misc/blocks-all.v1.yaml
+            minItems: 1
+        responsibilities:
```
thewilkybarkid: Wonder if it's simpler to just have a `content` property and these can be added as section blocks. They seem pretty granular. The set list of sections could be done at the input stage.

src/model/job-advert.v1.yaml
```diff
@@ -0,0 +1,45 @@
...
+        termsAndConditions:
+            $ref: ../blocks/paragraph.v1.yaml
+        offer:
+            description: Experience
+            type: array
+            items:
+                $ref: ../misc/blocks-all.v1.yaml
+            minItems: 1
+        contact:
+            description: Content
```
thewilkybarkid: What's this for?

src/model/job-advert.v1.yaml
```diff
@@ -0,0 +1,45 @@
...
+            type: string
+            format: email
+        closingDate:
```
thewilkybarkid: Probably should be in the snippet.

giorgiosironi: Is there a place where the designs are kept for the job adverts page?

src/model/job-advert.v1.yaml
giorgiosironi: For the email in `Please send your CV and a covering letter explaining your enthusiasm for this position and why you are a great person for this role to hr@elifesciences.org.` which seems pretty important to know when applying for a job

giorgiosironi: (the description is wrong)

src/model/job-advert.v1.yaml
giorgiosironi: There is a trade-off between a fixed schema and consistency. It seems most of the job adverts in the past cloned an existing one, tweaking the sections, but they were very similar for being generic blog posts. Still it's important to be able to customize in the future so some of these can be cut away in favor of the generic `content`, which we need anyway.

giorgiosironi:
This has been handed over to @davidcmoulton now.

src/model/job-advert.v1.yaml
davidcmoulton: Designs are in the ticket ELPP-1061.

src/model/job-advert.v1.yaml
thewilkybarkid: As far as I can tell this is always hr@elifesciences.org and is a form instead of that text (and gets shown elsewhere, eg 'No vacancies at present' page).

thewilkybarkid:
If we consider these like events (ie have a finite life), we should return all job adverts through the API, and have a filter to only return open/closed adverts.

davidcmoulton:
@thewilkybarkid Would the filtering be handled by `journal`?

thewilkybarkid:
Journal would use the query string parameter as it only shows open adverts, and would need to check the closing date when displaying the advert page.

Or, we really do consider the API as exposing open adverts only.

src/model/job-advert.v1.yaml
giorgiosironi: I think we should unify `description`, `responsibilities` and `experience` into the generic `content` mentioned, perhaps leave `termsAndConditions` as a separate element as it seems important/legal?

thewilkybarkid: If we ever have a use for it to be separate, we can extract it(/any others). But for now I don't see a reason to treat it differently.

giorgiosironi: :+1:

src/model/job-advert-list.v1.yaml
```diff
@@ -0,0 +1,17 @@
+$schema: http://json-schema.org/draft-04/schema#
+title: Job advert list
+type: object
+properties:
+    total:
+        title: Total number of open job adverts
```
thewilkybarkid: Remove 'open'.

src/samples/job-advert/v1/minimum.json
```diff
@@ -0,0 +1,15 @@
+{
+    "id": "80000099",
+    "title": "Editorial Assistant",
+    "impactStatement": "Support the Executive Editor in delivering an unparalleled editorial process and author experience",
+    "published": "2016-10-11T17:00:00Z",
+    "updated": "2016-10-11T17:00:00Z",
```
thewilkybarkid: Optional so shouldn't be here.

src/snippets/job-advert.v1.yaml
```diff
@@ -0,0 +1,29 @@
...
+        title: Publication date (UTC)
+        $ref: ../misc/date-time.v1.yaml
+    updated:
+        title: Updated date (UTC)
+        $ref: ../misc/date-time.v1.yaml
+
+required:
+  - id
+  - title
+  - impactStatement
```
thewilkybarkid: We tend to not make these required.

src/snippets/job-advert.v1.yaml
```diff
@@ -0,0 +1,28 @@
...
+        $ref: ../misc/date-time.v1.yaml
+    published:
+        title: Publication date (UTC)
+        $ref: ../misc/date-time.v1.yaml
+    updated:
+        title: Updated date (UTC)
+        $ref: ../misc/date-time.v1.yaml
+
+required:
+  - id
```
davidcmoulton: Make `impactStatement` required.

giorgiosironi:
Seems to be a pain to work with composer to include the branch and recompile it from the client perspective, especially in project that should have no knowledge of Node.js. Could we consider merging this annotating it as "not implemented yet"?

src/api.raml
```diff
@@ -799,6 +799,7 @@ traits:
                                     value: !include samples/interview/v1/complete.json
 
 /job-adverts:
+    (experimental):
```
thewilkybarkid: Need to add in https://github.com/elifesciences/api-raml/pull/192/files#diff-75d8606004fa61cbcecfcf88488241dbR70 too (this will get merged first).

giorgiosironi:
Can now be merged?

PR #179, opened 2017-08-02
ELPP-2601 Add support for thumbnail image for interviews

nlisgo:


src/snippets/interview.v1.yaml
```diff
@@ -20,6 +20,13 @@ properties:
     updated:
         title: Updated date (UTC)
         $ref: ../misc/date-time.v1.yaml
+    image:
+        type: object
+        properties:
+            thumbnail:
+                $ref: ../misc/image.v1.yaml
+        required:
+          - thumbnail
```
thewilkybarkid, line 29: Is the thumbnail required? If not, remove this.

nlisgo: It is my understanding that if we have an optional image then we must have a thumbnail to serve an image to the listing page.

Similar to the requirement for article listings: https://github.com/elifesciences/api-raml/blob/develop/src/snippets/article-vor.v1.yaml#L22

thewilkybarkid: Hmm, since `thumbnail` is overall optional there's no need to require `image` to contain a value if present. But, it doesn't make much of a difference and it already like this elsewhere, so let's leave it.

thewilkybarkid (approved):

PR #182, opened 2017-08-07
ELPP-2588 Add containing parameter to collections

thewilkybarkid:


src/api.raml
```diff
@@ -497,6 +497,13 @@ traits:
         is:
           - paged: { resources: Collections }
           - subjected: { resources: Collections }
+        queryParameters:
+            containing:
+                description: |
+                    Only return collections containing at least one of these items.
+                type: string[]
+                required: false
+                example: [article/1234, interview/5678]
```
thewilkybarkid, line 506: Slashes in URIs aren't ideal, maybe dashes (eg `article-1234`)?

giorgiosironi: also if the type is in the singular form, it gets confused with the URL `/articles/1234`. Ok for `article-1234`

thewilkybarkid: An eLife ID is `^[a-z0-9-]+$`, might be better to use something else as the delimiter? `article:1234-5678`?

thewilkybarkid: Will get confusing with something like `podcast-episode-4`.

thewilkybarkid: Recommendations and metrics have these are part of their URIs, but with slashes.

thewilkybarkid: See also https://github.com/elifesciences/api-client-php/pull/67

giorgiosironi: Recommendations and metrics use these values, but as part of the path, not of the query string. `:` is also unambiguous and would work.
In builder we use `--` (which has its own problems like being recognized as ~strikethrough~ by Jira) but only because other characters are not allowed.

thewilkybarkid: I like the idea of a consistent way to refer to a content type + its ID. So using colons would mean small changes to Recommendations and Metrics (to support both, and deprecate using slashes). /cc @lsh-0

giorgiosironi: Let's wait for @lsh-0 comment and if we go ahead with `:` I'll open some PRs.

lsh-0: I've always been suspicious of pushing unrelated changes through for the sake of consistency in an awkward and inconsistent bit of the program. It strikes me that what collections needs here (a list of id and type pairs) feels awkward because it's not in keeping with the rest of the api, and that there are sort-of similar cases where a type is supplied along with an id in the path, like `/article/09560/page-views`. And you're suggesting to make it `/page-views?foo=article:09560`, right? or is it `/article:09560/page-views` ?

Either way, it seems like you're trying to backport an awkward design decision in one app in one case to multiple existing apps and future apps for all cases, just for consistency's sake. This ignores that metrics doesn't support multiple types and ids in one request.

`/collections?containing=article:09560&containing=interview:12345` seems fine to me. As for delimiters, I prefer the pipe `|` but I really don't care in this case. 

When we have two such instances where multiple types and ids must be passed, we could look at it again then and see how well collections fared using whatever decision you make now.

:-1:

giorgiosironi: Using the differentiation "one possible item" (`/article/09560`, recommendations and metrics) vs. "multiple possible items" (`?containing[]=article:09560&containing[]=article:09561`, journal-cms) could be a good solution then.

lsh-0: I think that's an important distinction.

:+1:

thewilkybarkid: I think I over-thought this (when I recognised https://github.com/elifesciences/api-client-php/pull/67 it simplified things, as the slashes don't have to be percent-encoded).

Since PRs are already open, I'm happy to stick with slashes here. Agree than 1 vs many is path vs query string parameters.

giorgiosironi (approved):

thewilkybarkid:
Will merge when the implementation is ready.

PR #185, opened 2017-08-17
Podcast episodes containing

thewilkybarkid:
Allows https://github.com/elifesciences/recommendations/pull/83 to include podcast episodes.

(Technically it wants the _chapters_, but until we add an endpoint for chapters this will do.)

### DIFF
--- a/src/api.raml
+++ b/src/api.raml
@@ -176,7 +176,7 @@ traits:
                   - published
                 required: false
                 default: default
-    contained:
+    container:
         queryParameters:
           containing[]:
               description: |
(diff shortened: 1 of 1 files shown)


src/api.raml
```diff
@@ -176,6 +176,14 @@ traits:
                   - published
                 required: false
                 default: default
+    contained:
```
giorgiosironi: strange name `contained` since this item is the one that contains others (maybe `container`)

giorgiosironi (approved):

PR #186, opened 2017-09-11
PI-499 Make label required and title optional in asset-file

thewilkybarkid:
This is a subtle change that has quite a large visual impact.

In `asset-file` (use by figure source data and article additional files) `label` should have been required and `title` optional.


src/blocks/section.v2.yaml
```diff
@@ -0,0 +1,24 @@
+$schema: http://json-schema.org/draft-04/schema#
```
giorgiosironi, line 1: are we making new versions of all these blocks because `asset-file` is used in `figure` and eventually in these content types?

thewilkybarkid: Yep. 😒 

(Sharing the same structure across all the content types is great for consumption as they're all the same. And then painful for adding new/changing existing types...)

lsh-0:
```
[ml_PR-186-T56L6E4MNFVDBBRZCMYI] Running shell script
+ git fetch
+ git diff --exit-code origin/develop...HEAD dist/
fatal: ambiguous argument 'origin/develop...HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
script returned exit code 128
```

lsh-0:
in `dist/model/` should I expect to have a `article-vor.v2.json` and `article-poa.v2.json` files after using npm to compile? I'm trying to get lax to validate against one or the other

src/misc/article.v2.yaml
```diff
@@ -0,0 +1,78 @@
...
+            items:
+                $ref: ../blocks/paragraph.v1.yaml
+            minItems: 1
+        funding:
+            $ref: ../misc/funding.v1.yaml
+        additionalFiles:
+            title: Additional files
+            type: array
+            items:
+                $ref: ../misc/asset-file.v1.yaml
```
gnott: Could you please take a look here and confirm - perhaps by @thewilkybarkid - should this be ``asset-file.v1.yaml`` or should it be ``asset-file.v2.yaml``?

I'm testing the schema generating an altered article 26759, except it is not passing validation because the transparent reporting form in the additionalFiles has a label and not a title.

Also above in ``article.v1.yaml`` it is referencing to use ``asset-file.v2.yaml``.

I wanted to confirm these before I presumably change how the title or label is set depending on whether it is a figure sourceData or an article additionalFiles item (which I guess in the XML parser goes through the same function).

thewilkybarkid: Good spot, I've just found a bunch of these. :doh:

thewilkybarkid:
Could mark the new versions as experimental and merge?

giorgiosironi:
:+1: from me, I guess the annotation should be the first field of the v2 schema definition?

thewilkybarkid:
Yeah, like https://github.com/elifesciences/api-raml/pull/191/files#diff-75d8606004fa61cbcecfcf88488241dbR1145

src/blocks/section.v2.yaml
giorgiosironi: Good old http://wiki.c2.com/?StableDependenciesPrinciple

giorgiosironi:
`curl -v -H "Accept: application/vnd.elife.blog-article+json; version=2" https://prod--gateway.elifesciences.org/blog-articles/f1ca9823` fails at the moment, so also the journal-cms ones have to be marked as `(experimental):`

giorgiosironi:
Tests pass, marked, merging.

PR #188, opened 2017-09-15
Add schema for annotations

nlisgo:


src/model/annotation-list.v1.yaml
```diff
@@ -0,0 +1,49 @@
...
+        title: Total number of Annotations
+        type: integer
+        minimum: 0
+    items:
+        title: Annotationss
+        type: array
+        items:
+            type: object
+            properties:
+                uri:
```
giorgiosironi: does the `uri` work as an id? To uniquely characterize an annotation

src/model/annotation-list.v1.yaml
```diff
@@ -0,0 +1,49 @@
...
+                    description: Publication date (UTC)
+                    $ref: ../misc/date-time.v1.yaml
+                updated:
+                    title: Updated date (UTC)
+                    $ref: ../misc/date-time.v1.yaml
+                content:
+                    title: Content
+                    type: array
+                    items:
+                        oneOf:
```
giorgiosironi, line 51: what's the rationale for including some types of blocks but not others? a comment or a link could help explain that

src/api.raml
```diff
@@ -185,6 +185,37 @@ traits:
               required: false
               example: [article/1234, interview/5678]
 
+/annotations/{id}:
+    type: collection
+    description: |
+        Annotations
+    uriParameters:
+        id:
```
thewilkybarkid: That looks like the ID of an annotation. Instead, maybe we should have `by` as a query string parameter, and make it required for now (so if we ever allow lists by everyone, make it optional). (Or even by multiple authors, allow it to be multiple.)

src/model/annotation-list.v1.yaml
```diff
@@ -0,0 +1,49 @@
+$schema: http://json-schema.org/draft-04/schema#
+title: Annotation list
+type: object
+properties:
+    total:
+        title: Total number of Annotations
+        type: integer
+        minimum: 0
+    items:
+        title: Annotationss
```
thewilkybarkid: Typo

src/model/annotation-list.v1.yaml
```diff
@@ -0,0 +1,49 @@
...
+                uri:
+                    description: Annotation URI
+                    type: string
+                    format: uri
+                title:
+                    description: Title
+                    type: string
+                target:
+                    description: Target text for the annotation
+                    type: string
```
thewilkybarkid: Will this ever contain HTML?

src/model/annotation-list.v1.yaml
```diff
@@ -0,0 +1,49 @@
...
+        type: array
+        items:
+            type: object
+            properties:
+                uri:
+                    description: Annotation URI
+                    type: string
+                    format: uri
+                title:
+                    description: Title
```
thewilkybarkid: Title of what? The annotation?

src/model/annotation-list.v1.yaml
```diff
@@ -0,0 +1,49 @@
...
+                content:
+                    title: Content
+                    type: array
+                    items:
+                        oneOf:
+                          - $ref: ../blocks/list.v1.yaml
+                          - $ref: ../blocks/mathml.v1.yaml
+                          - $ref: ../blocks/paragraph.v1.yaml
+                          - $ref: ../blocks/quote.v1.yaml
+                          - $ref: ../blocks/youtube.v1.yaml
```
thewilkybarkid: `minItems: 1`?

src/samples/annotation-list/v1/first-page.json
```diff
@@ -0,0 +1,47 @@
...
+                    "text": [
+                        {
+                            "type": "paragraph",
+                            "text": "Blockquote"
+                        }
+                    ]
+                },
+                {
+                    "type": "paragraph",
+                    "text": "https://cdn.elifesciences.org/style-guide-images/elife-full-color-horizontal.png"
```
thewilkybarkid, line 34: Is `image` meant to be a possibility?

src/model/annotation-list.v1.yaml
nlisgo: I will check but I believe that hypothesis strip out the HTML.

src/api.raml
nlisgo: Will make this adjustment.

src/model/annotation-list.v1.yaml
nlisgo: This would be the title of the page that has been annotated.

src/samples/annotation-list/v1/first-page.json
nlisgo: This could be interesting. The images that are displayed in the annotations are only referenced, they must be hosted elsewhere. We only support displaying images served on a IIIF server. I will talk to @NickDuf about how we should present these on the profiles page.

src/model/annotation-list.v1.yaml
nlisgo: `minItems: 1` has been added.

src/samples/annotation-list/v1/first-page.json
giorgiosironi: For what it's worth, it's insanely easy to add a new source to our IIIF that points to `cdn.elifesciences.org` or to somewhere similar.

thewilkybarkid: So the images aren't hosted by Hypothesis? We might want to do what we do with Medium and download them, and connect Loris to Annotations. (Preserves the record of the image too, though I guess it would change/disappear through the Hypothesis client?)

nlisgo: Yes, the hypothesis client would not use our images unless we edited the annotation and that would not be appropriate. The images in an annotation are stored as url's only. They could be hosted absolutely anywhere.

My recommendation for now is that we either present them as clickable links or that we convert them to ``. Will chat to @NickDuf.

src/model/annotation-list.v1.yaml
nlisgo: I believe that these will be the only blocks needed to present all of the content for the annotations. This is due to the limited markup that is permitted within an annotation on hypothesis. As this is still experimental we could revise it once we have migrated all of the annotations and we know whether we should be supporting any other blocks.

src/model/annotation-list.v1.yaml
nlisgo: It will most certainly be unique. Here is an example:

```
https://hyp.is/36GLapojEeektWtTtv2WnA/elifesciences.org/inside-elife/77dd2808/taking-a-closer-look-screening-of-images
```
The hypothesis annotation id is contained within the url `36GLapojEeektWtTtv2WnA`.

src/model/annotation-list.v1.yaml
```diff
@@ -0,0 +1,50 @@
...
+                    minItems: 1
+            required:
+              - uri
+              - title
+              - published
+              - content
```
nlisgo: I think I may need to make content optional because we may wish to displayed highlighted text without an annotation.

src/model/annotation-list.v1.yaml
thewilkybarkid: Hmm, the slug shouldn't be part of the ID as renaming the item causes it to change. Can we declare our ID, which would be `blog-article/77dd2808` in this case?

src/samples/annotation-list/v1/first-page.json
thewilkybarkid: We already handle inline images, so that would work. But if they include something that's massive/disappears...

src/model/annotation-list.v1.yaml
thewilkybarkid: Ah, forgot that was a thing. Need to revert `content.minItems` as well.

src/model/annotation-list.v1.yaml
```diff
@@ -0,0 +1,50 @@
...
+        type: array
+        items:
+            type: object
+            properties:
+                uri:
+                    description: Annotation URI
+                    type: string
+                    format: uri
+                title:
+                    description: Title of the annotated document
```
thewilkybarkid: As in the article title? We might want its type+ID too? Maybe we should include the snippet instead.

src/model/annotation-list.v1.yaml
nlisgo: I agree. I have not introduced an id, I was unclear whether it was necessary since I am not exposing an endpoint to retrieve a single annotation.

src/model/annotation-list.v1.yaml
nlisgo: It would be the title of whichever document that has been annotated: Article, blog article, labs post.

If we were to serve up the snippet of the content then this would require us to derive that from the target url which shouldn't be too much trouble but would require us to do a request for each annotation to the relevant service or to store the snippets for all content as part of the annotations service.

Having reviewed the designs for the annotations to be displayed on the profiles page, adding the snippets into the response would not add much value at this stage other than ensuring that the correct title of the annotated document was displayed. As things stand it would be the title of the document at the time that the annotation was recorded.

src/model/annotation-list.v1.yaml
nlisgo: As content is optional, is it not ok to leave the `minItems` as `1`. It is my understanding that you could omit the property entirely but if it is present then it should contain at least `1` item.

src/samples/annotation-list/v1/first-page.json
nlisgo: @thewilkybarkid we cannot control what the size of the image is that is referenced. The hypothes.is client will attempt to render it regardless.

src/model/annotation-list.v1.yaml
thewilkybarkid: > As things stand it would be the title of the document at the time that the annotation was recorded.

Interesting, so different to the snippet then. We can ignore that for now, and add later if we need it.

src/model/annotation-list.v1.yaml
thewilkybarkid: Easier to make empty the same as not there.

src/model/annotation-list.v1.yaml
thewilkybarkid: Sounds like we don't want to use their URI at all, instead we can change this to an ID property.

src/samples/annotation-list/v1/first-page.json
thewilkybarkid: Depending on what the design request is, this should become HTML (eg `` or `<a>`). Maybe one day we can integrate Hypothesis and IIIF...

src/model/annotation-list.v1.yaml
nlisgo: It has been removed again as content is optional.

src/samples/annotation-list/v1/first-page.json
nlisgo: At the moment the designs are showing it displayed as a text link `<a>`.

src/model/annotation-list.v1.yaml
```diff
@@ -0,0 +1,62 @@
...
+        items:
+            type: object
+            properties:
+                id:
+                    description: Annotation ID
+                    type: string
+                private:
+                    description: If the annotation should only be visible to the author
+                    type: boolean
+                    default: false
```
thewilkybarkid: We should match what we did for profiles https://github.com/elifesciences/api-raml/blob/d2c08de4c7f3493cb8bbed9200b23f2cab0da226/src/model/profile.v1.yaml#L14-L15

src/model/annotation-list.v1.yaml
```diff
@@ -0,0 +1,62 @@
...
+                    $ref: ../misc/date-time.v1.yaml
+                updated:
+                    title: Updated date (UTC)
+                    $ref: ../misc/date-time.v1.yaml
+                content:
+                    title: Content
+                    type: array
+                    items:
+                        oneOf:
+                          - $ref: ../blocks/list.v1.yaml
```
thewilkybarkid: `v2` (not that we see the stuff that changed here)

src/model/annotation-list.v1.yaml
```diff
@@ -0,0 +1,62 @@
...
+                    type: boolean
+                    default: false
+                title:
+                    description: Title of the annotated document
+                    type: string
+                target:
+                    description: Target text for the annotation
+                    type: string
+                targetUri:
+                    description: URI of the annotated document
```
thewilkybarkid: https://github.com/elifesciences/api-raml/pull/188#discussion_r146541874 is still relevant here.

src/model/annotation-list.v1.yaml
```diff
@@ -0,0 +1,89 @@
...
+                    type: array
+                    items:
+                        oneOf:
+                          - $ref: ../blocks/list.v2.yaml
+                          - $ref: ../blocks/mathml.v1.yaml
+                          - $ref: ../blocks/paragraph.v1.yaml
+                          - $ref: ../blocks/quote.v1.yaml
+                          - $ref: ../blocks/youtube.v1.yaml
+                item:
+                    oneOf:
```
nlisgo: in order to introduce these snippets we are going to have to store them as part of the annotations service or perform an API request for each annotation, so we might need to rethink this? @giorgiosironi @thewilkybarkid

giorgiosironi: How many annotations are we talking about in a single response?

Recommendations uses the approach of making several internal API requests, without cache, to compose a result. However, they are not that many, ~5 requests per API response tops.
Search is doing the same but with caching (I think?) or is it actually persisting in ElasticSearch the snippets? Will check.

Just to understand the whole context, where does the requirement of serving the snippets came from?

nlisgo: Each annotation will have a snippet. So, if you request 20 annotations in a list then there will be up to 20 requests.

giorgiosironi: Search is not caching any API call explicitly, the snippets are stored in ElasticSearch and overwritten on bus events. Which is what `annotations` would have to do, with a different persistence storage probably.

nlisgo: We can get the title and url of the highlighted article and the Hypothesis API response which is all we need for the display but it doesn't seem right not to include the whole snippet.

src/model/annotation-list.v1.yaml
```diff
@@ -0,0 +1,65 @@
...
+                    type: string
+                access:
+                    $ref: ../misc/access-level.v1.yaml
+                hasParent:
+                    description: If the annotation has a parent annotation
+                    type: boolean
+                    default: false
+                document:
+                    title: Annotated document
+                    type: object
```
nlisgo: This object is made up of values already available in a hypothesis response. It should give us everything we need to serve the designs.

src/model/annotation-list.v1.yaml
```diff
@@ -0,0 +1,65 @@
...
+        type: array
+        items:
+            type: object
+            properties:
+                id:
+                    description: Annotation ID
+                    type: string
+                access:
+                    $ref: ../misc/access-level.v1.yaml
+                hasParent:
```
giorgiosironi: is there a way to find out the `id` or something similar for the parent annotations? Rather than only knowing if it has one? This to promote reusability of this API with a low effort.

giorgiosironi: All the other field names seem to match the other schemas in this repository

nlisgo: In the annotation schema the `references` property lists the annotation id of all parents.

https://github.com/hypothesis/h/blob/75714a4829bf17cdb760b98829942b2646e60542/docs/_extra/api-reference/schemas/annotation-schema.json#L25-L30

I will introduce this array. The last entry in the array is the immediate parent and the first entry is the top level annotation or page note. If it is empty, it has no parents and is a top-level annotation or page note.

nlisgo:
One thing that belongs here also is the profile id and display name of the author of the annotation. These values are returned in the Hypothesis search response but we want to avoid fleshing out the profile snippet because that will potentially require an extra API call for each annotation.

giorgiosironi:
Makes sense given the naming `parents`, although an empty array is usually the "NULL value" of lists

giorgiosironi (approved):
I have no further concerns

src/model/annotation-list.v1.yaml
```diff
@@ -0,0 +1,66 @@
...
+        items:
+            type: object
+            properties:
+                id:
+                    description: Annotation ID
+                    type: string
+                access:
+                    $ref: ../misc/access-level.v1.yaml
+                parents:
+                    description: Array of parent annotation ID's
```
thewilkybarkid: `Array of` is redundant; no need for an apostrophe.

src/model/annotation-list.v1.yaml
```diff
@@ -0,0 +1,66 @@
...
+                          - $ref: ../blocks/paragraph.v1.yaml
+                          - $ref: ../blocks/quote.v1.yaml
+                          - $ref: ../blocks/youtube.v1.yaml
+            required:
+              - id
+              - access
+              - parents
+              - document
+              - published
+              - updated
```
thewilkybarkid: Not required?

src/model/annotation-list.v1.yaml
```diff
@@ -0,0 +1,66 @@
...
+                            type: string
+                            format: uri
+                    required:
+                      - title
+                      - uri
+                highlight:
+                    description: Highlighted text
+                    type: string
+                published:
+                    description: Publication date (UTC)
```
thewilkybarkid: Of the annotation? Probably need a different term here.

src/model/annotation-list.v1.yaml
```diff
@@ -0,0 +1,66 @@
...
+                    description: Annotation ID
+                    type: string
+                access:
+                    $ref: ../misc/access-level.v1.yaml
+                parents:
+                    description: Array of parent annotation ID's
+                    type: array
+                    items:
+                        type: string
+                document:
```
thewilkybarkid: Not a fan of these, would much rather this API translated it into our concepts (in this case content item). But it does allow us to do what we need (and redirects should cover any changes), so I can live with it (for now).

src/api.raml
```diff
@@ -191,6 +191,40 @@ traits:
...
+        Annotations
+    get:
+        description: |
+            Get a list of annotations.
+        is:
+          - paged: { resources: Annotations }
+        queryParameters:
+            by[]:
+                description: |
+                    Only return annotations by at least one of these authors.
```
thewilkybarkid: As we're not going to be including author details, we should probably make this a single author, or it's a bit confusing?

src/api.raml
```diff
@@ -191,6 +191,51 @@ traits:
...
+        description: |
+            Get a list of annotations, sorted by date.
+        is:
+          - paged: { resources: Annotations }
+        queryParameters:
+            by:
+                description: |
+                    Only return annotations by this author.
+                type: string[]
+                required: false
```
nlisgo: I've set this to be optional now.

src/model/annotation-list.v1.yaml
```diff
@@ -0,0 +1,66 @@
...
+            type: object
+            properties:
+                id:
+                    description: Annotation ID
+                    type: string
+                access:
+                    $ref: ../misc/access-level.v1.yaml
+                parents:
+                    description: Parent annotation IDs
+                    type: array
```
nlisgo, line 22: Could this just be `type: string[]`?

thewilkybarkid: No (this is JSON Schema, not RAML).

src/api.raml
thewilkybarkid: I didn't think we were going to list _all_ annotations? This has the same author problem (you don't know whose annotation it is).

nlisgo: Right, I'll set it back to required.

thewilkybarkid (approved):

PR #189, opened 2017-09-15
[WIP] Create schema for annotation token

nlisgo:


src/api.raml
```diff
@@ -185,6 +185,28 @@ traits:
...
+            type: string
+    get:
+        description: |
+            Get an annotation token.
+        responses:
+            200:
+                description: |
+                    Annotation token.
+                body:
+                    application/vnd.elife.annotation-token+json;version=1:
```
giorgiosironi, line 205: I wonder if there is a standard way to mark this as experimental as we will need to merge to go live, and do breaking changes later. Perhaps there is a standard field in RAML? Or could we use `version=0`?

src/misc/jwt.v1.yaml
```diff
@@ -0,0 +1,5 @@
+$schema: http://json-schema.org/draft-04/schema#
+
+title: Token
+type: string
```
giorgiosironi: link to where the regex comes from?
and is it worth it to lock down?

src/model/annotation-token.v1.yaml
```diff
@@ -0,0 +1,12 @@
+$schema: http://json-schema.org/draft-04/schema#
+title: Annotation token
+type: object
+allOf:
+  - properties:
+        token:
+            $ref: ../misc/jwt.v1.yaml
+        expire:
```
giorgiosironi: we've seen in `profiles` `expires_at` to distuinguish between moments in time and intervals (which would be `expires_in`). Perhaps here too?

nlisgo: Will change to `expiresAt`

src/misc/jwt.v1.yaml
nlisgo: I have found a better regex: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/master/src/System.IdentityModel.Tokens.Jwt/JwtConstants.cs#L58

src/api.raml
nlisgo: Let's discuss this with @thewilkybarkid.

thewilkybarkid: RAML has annotations. Thinking about it, we should do that for deprecations too. I'll take a look at adding them.

src/samples/annotation-token/v1/complete.json
```diff
@@ -0,0 +1,4 @@
+{
+    "token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJhdWQiOiJoeXBvdGhlcy5pcyIsImlzcyI6ImNsaWVudF9pZCIsInN1YiI6ImFjY3Q6dXNlckBhdXRob3JpdHkiLCJuYmYiOjE1MDU0OTU4NzgsImV4cCI6MTUwNTQ5NjQ3OH0.FttNDnvdXMeSVaKVDzoYUW7RsP7ovMVwNmNRVZW8OsE",
+    "expire": "2017-09-15T19:27:58Z"
```
thewilkybarkid, line 3: `expiresAt`?

src/api.raml
thewilkybarkid: We've started using them, so the whole API should be `@experimental`.

thewilkybarkid:
This isn't needed.

PR #191, opened 2017-09-20
Add RAML annotations

thewilkybarkid:


thewilkybarkid:
API Console has no support for displaying annotations.

src/api.raml
```diff
@@ -1138,6 +1142,7 @@ traits:
                         Press package.
                     body:
                         application/vnd.elife.press-package+json;version=1:
+                            (deprecated):
```
thewilkybarkid, line 1329: I _think_ this is valid RAML, but the JS library doesn't recognise it. (Making it a require string doesn't throw an error, but does if you move it directly inside `body`.)

giorgiosironi: seems valid according to https://raml.org/developers/whats-new-raml-10 although they use `null` rather than `nil`

thewilkybarkid: They changed it to `nil` to avoid confusion with YAML's `null` type (https://github.com/raml-org/raml-spec/issues/514).

thewilkybarkid:
We should add an ADR for the `deprecated` and `experimental` annotations. /cc @giorgiosironi

giorgiosironi:
Agreed, I'll do it.

adr/0003-use-raml-annotation-deprecated.md
```diff
@@ -0,0 +1,23 @@
+# 2. Use experimental annotation
```
thewilkybarkid: Deprecated

adr/0003-use-raml-annotation-deprecated.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+Versioned APIs need to retire older versions of a schema, in order to keep complexity manageable.
+
+It is not possible to deploy different services atomically so that a new version can be switched on in a short time frame. The changes needed have to be propagated to both clients and servers.
```
thewilkybarkid, line 13: Maybe add that we already add a `Warning` header on responses.

adr/0003-use-raml-annotation-deprecated.md
```diff
@@ -0,0 +1,25 @@
...
+
+We add `Warning` headers to responses that are using an old version.
+
+## Decision
+
+We will annotate with `@deprecated` API versions that are not served by default anymore, whenever a new version is added.
```
thewilkybarkid: As long as the new version isn't `@experimental`?

PR #193, opened 2017-10-09
ELPP-2814 Add profile block

thewilkybarkid:
Adding to the new versions caused by #186.


giorgiosironi (approved):
Doesn't create complexity on the `article` change (still getting deployed) so should be ok. IIRC all new versions were tagged as @experimental

thewilkybarkid:
@nlisgo Primarily for blog articles.

PR #194, opened 2017-10-17
proposal for journal page view counts

lsh-0:
I needed something to work with. 

I figured each page would be asking for it's own views and there would just be a single result.

I also added in a `target` uri param should we ever want to start capturing metrics for other non-journal targets (maybe crm or annual report mini-sites?)

Feel free to completely rework it if need be.


thewilkybarkid (changes requested):
We need the same behaviour as articles, that is store the history of item across all the URIs that it's ever had (and we have no requirement for counting non-content item pages). https://github.com/elifesciences/api-raml/blob/develop/src/api.raml#L1109-L1110 should be expanded to include the new types and that API reused instead (just with no time periods). (`/downloads` should probably 404?)

lsh-0:
~~I'd be grateful if you could modify this to how you think it should look.~~

I've updated the ticket, once I get some clarity on where this feature is going I'll update the api again

lsh-0:
~fixed up, @thewilkybarkid.~ urgh, ignore me. not ready.

I'm getting confused by what is considered a `type` and what is an `id` here.

In the case I have here, I have `type` as `page` and `id` as a list of known page types that all paths returned by google analytics will fit under. This doesn't account for wanting to know the view count for a page at `/inside-elife/2342341/foo-bar` for example.

I have the strong suspicion that what you actually want is `type` to be a list of known api-raml types (like collection, event, interview) and the id to be that bit of content's `id` value, and that specifying the path to fetch for this endpoint wouldn't be necessary ... I'm just not sure how strong the relationship is between the journal routing and the api routing and the set of objects you're using.

If you have a master plan here it might be best you fix up what I've done to match what you're after.

PR #195, opened 2017-10-17
Make api-raml JSON schemas available as python package

seanwiseman:
I have added the required changes to allowing the following functionality to be possible:

`pip install -e git+https://github.com/elifesciences/api-raml.git#egg=api-raml`

```
from py_api_raml import schema_directory
import os

>>> print(schema_directory)
'/Users/someuser/somefolder/api-raml/venv/lib/python3.5/site-packages/py_api_raml-0.0.1-py3.5.egg/py_api_raml'

>>> os.listdir(schema_directory)
['article-poa.v1.json', 'press-package.v3.json', 'article-related.v1.json'.......]
```

The 'schema_directory' string allows you to view the contents of the directory and to also perform actions on the files based on the fact you have their directory path. e.g.

```
file_path = os.path.join(schema_directory, 'article-pop.v1.json')
with open(file_path) as f:
    data = json.load(f)
```

At this point I think keeping it off PyPi will allow for more flexibility. We can still list it as a requirement via its git address and pip will handle it fine. If we created it as a PyPi package we would need to build and re version after each change/build of api-raml.

### DIFF
--- a/compile.js
+++ b/compile.js
@@ -7,13 +7,12 @@ var recursive = require('recursive-readdir');
 var refParser = require('json-schema-ref-parser');
 
 fs.removeSync('./dist');
-fs.removeSync('./py_api_raml'); // for py
+fs.removeSync('./elife_api');
 
 fs.mkdirsSync('./dist/model');
 fs.mkdirsSync('./dist/samples');
 fs.mkdirsSync('./dist/styles');
-
-fs.mkdirsSync('./py_api_raml'); // for py
+fs.mkdirsSync('./elife_api');
 
 recursive('./src/model', function (err, files) {
     var promises = [];
@@ -68,8 +67,6 @@ recursive('./src/model', function (err, files) {
         fs.copy('./src/styles', './dist/styles');
         fs.copy('./src/index.html', './dist/index.html');
         fs.copy('./src/favicon.ico', './dist/favicon.ico');
-
-        fs.copy('./src/__init__.py_base', './py_api_raml/__init__.py'); // for py
-        fs.copy('./dist/model', './py_api_raml'); // for py
+        fs.copy('./src/__init__.py_base', './elife_api/__init__.py');
     });
 });
--- a/compile.js
+++ b/compile.js
@@ -7,9 +7,13 @@ var recursive = require('recursive-readdir');
 var refParser = require('json-schema-ref-parser');
 
 fs.removeSync('./dist');
+fs.removeSync('./py_api_raml'); // for py
+
 fs.mkdirsSync('./dist/model');
 fs.mkdirsSync('./dist/samples');
 fs.mkdirsSync('./dist/styles');
+
+fs.mkdirsSync('./py_api_raml'); // for py
 
 recursive('./src/model', function (err, files) {
     var promises = [];
@@ -64,5 +68,8 @@ recursive('./src/model', function (err, files) {
         fs.copy('./src/styles', './dist/styles');
         fs.copy('./src/index.html', './dist/index.html');
         fs.copy('./src/favicon.ico', './dist/favicon.ico');
+
+        fs.copy('./src/__init__.py_base', './py_api_raml/__init__.py'); // for py
+        fs.copy('./dist/model', './py_api_raml'); // for py
     });
 });
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ import elife_api
 setup(
     name='elife_api',
     version=elife_api.__version__,
-    description='Utility for accessing eLife API-Raml JSON schemas in Python.',
+    description='Utility for accessing eLife API JSON Schemas.',
     packages=['elife_api'],
     license='MIT',
     url='https://github.com/elifesciences/api-raml',
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,22 @@
+from setuptools import setup
+
+import elife_api
+
+setup(
+    name='elife_api',
+    version=elife_api.__version__,
+    description='Utility for accessing eLife API-Raml JSON schemas in Python.',
+    packages=['elife_api'],
+    license='MIT',
+    url='https://github.com/elifesciences/api-raml',
+    maintainer='eLife Sciences Publications Ltd.',
+    maintainer_email='py@elifesciences.org',
+    classifiers=[
+        "Development Status :: 1 - Alpha",
+        "Intended Audience :: Developers",
+        "License :: OSI Approved :: MIT License",
+        "Programming Language :: Python :: 2.7",
+        "Programming Language :: Python :: 3.5",
+        ],
+    include_package_data=True
+)
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,7 @@ setup(
     license='MIT',
     url='https://github.com/elifesciences/api-raml',
     maintainer='eLife Sciences Publications Ltd.',
-    maintainer_email='py@elifesciences.org',
+    maintainer_email='tech-team@elifesciences.org',
     classifiers=[
         "Development Status :: 1 - Alpha",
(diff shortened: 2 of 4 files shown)


thewilkybarkid (commented):
What's the impact of putting `__init__.py` inside `dist/model` rather than creating another copy of the compiled files? The module name is then `model`? Can it be aliased somehow?

src/__init__.py_base
```diff
@@ -0,0 +1,11 @@
+import os
+
+import py_api_raml
```
thewilkybarkid: We called the PHP package `elife/api`, so maybe we should aim for `elife_api`?

giorgiosironi:
We use the "point to a Git sha and skip PyPI" method: https://github.com/elifesciences/bot-lax-adaptor/blob/develop/requirements.txt#L5 so no problem with that. One thing that bit us was the downstream project not updating the virtualenv with the new version when the sha was changed in `requirements.txt`, because the version was the same. Perhaps we are doing the update incorrectly, check https://github.com/elifesciences/bot-lax-adaptor/blob/develop/install.sh#L17-L19 is actually needed.

compile.js
```diff
@@ -64,5 +68,8 @@ recursive('./src/model', function (err, files) {
         fs.copy('./src/styles', './dist/styles');
         fs.copy('./src/index.html', './dist/index.html');
         fs.copy('./src/favicon.ico', './dist/favicon.ico');
+
+        fs.copy('./src/__init__.py_base', './py_api_raml/__init__.py'); // for py
+        fs.copy('./dist/model', './py_api_raml'); // for py
```
giorgiosironi: to be aware of: once something is committed to `develop`, https://ci--alfred.elifesciences.org/job/library-api-raml/ builds it and commits the changes to `dist` on top of it. That's the commit that should be used by downstream projects.
Perhaps we need to add more entries to https://github.com/elifesciences/api-raml/blob/develop/Jenkinsfile#L19?

src/__init__.py_base
seanwiseman: Fine with the name change.

I did try putting it in the dist (it does work from there) but the only issue was that I was to forced to then import it as:
`from dist import schema_directory`

As I could not rename the actual 'dist' directory to which the module name is pulled from I added that a python specific directory be generated during the build. But yes it would nicer to not have copied files, but I think having to `from dist import ..` is too confusing.

compile.js
giorgiosironi: Definitely needs to be added then: `def isThereANewCommit = elifeGitAutoCommit "Regenerated dist/ and dependencies", "dist/", "py_api_raml/`

src/__init__.py_base
giorgiosironi: Downside is that the repository will contain 3 copies (`src`, `dist` and `py`) of schemas, which can be also confusing when using grep or searches. Is it possible to commit the `py_api_raml` module where it should be, and do magic inside it (realpath and `../`) to get to the `dist/` folder?

compile.js
seanwiseman: Agreed. I will change the name to `elife_api` as suggested by @thewilkybarkid and then add the  change to the Jenkinsfile.

src/__init__.py_base
seanwiseman: OK I have got around it. Best of both worlds I believe, the module `elife_api` will now only contain its `__init__.py` and I just changed the `include` path in the `MAINFEST.in` to be `include dist/model/*.json`. This way the JSON files are not duplicated, just pulled into the egg when someone installs the package, and we also don't have to call the module `dist`.

compile.js
seanwiseman: Now that I am not storing the JSON copies in the `elife_api` directory I don't think this is required.

giorgiosironi: Correct, as the code added here won't be affected by the regeneration

seanwiseman: Regarding https://github.com/elifesciences/bot-lax-adaptor/blob/develop/install.sh#L17-L19 I think something along these lines would be needed by projects dependant on it in order to keep `elife_api` up to date. If the version value in the `setup.py` does not change then pip will skip it by default.

The alternative is to use the `--upgrade --force-reinstall` args when running pip, this will reinstall even if the version has not changed. The downside being that if we did this in line with `pip install -r requirements.txt` everything will get reinstalled.

giorgiosironi: Correct to keep the `elife_api` version pinned, every build starting from the same project commit should get the same result.
The pattern we use for dependency updates is to automate opening a pull request upon certain events - every week or when the upstream gets a new commit if it's something that must be kept up to date.

See:
- https://github.com/elifesciences/bot-lax-adaptor/pulls?utf8=%E2%9C%93&q=is%3Apr%20api-raml%20 (sample api-raml proposed updates in bot-lax-adaptor)
- https://github.com/elifesciences/bot-lax-adaptor/blob/develop/Jenkinsfile.update-api-raml (different builds that is triggered to do the update)
- https://github.com/elifesciences/bot-lax-adaptor/blob/develop/update-api-raml.sh weird script that does what pip should be doing. We would run `--upgrade` in a separate build which will open a PR rather than run tests.

setup.py
```diff
@@ -0,0 +1,22 @@
...
+
+setup(
+    name='elife_api',
+    version=elife_api.__version__,
+    description='Utility for accessing eLife API-Raml JSON schemas in Python.',
+    packages=['elife_api'],
+    license='MIT',
+    url='https://github.com/elifesciences/api-raml',
+    maintainer='eLife Sciences Publications Ltd.',
+    maintainer_email='py@elifesciences.org',
```
thewilkybarkid: Does that exist? Could be `tech-team@elifesciences.org`?

setup.py
```diff
@@ -0,0 +1,22 @@
+from setuptools import setup
+
+import elife_api
+
+setup(
+    name='elife_api',
+    version=elife_api.__version__,
+    description='Utility for accessing eLife API-Raml JSON schemas in Python.',
```
thewilkybarkid: Maybe just `Utility for accessing eLife API JSON Schemas.`?

src/__init__.py_base
```diff
@@ -0,0 +1,11 @@
+import os
+
+import elife_api
+
+__version__ = '0.0.1'
```
thewilkybarkid, line 5: We haven't versioned this repo, though I think we could.

setup.py
seanwiseman: Ok changed.

setup.py
seanwiseman: I got it from https://github.com/elifesciences/elife-tools/blob/05def8780aaf7ffb247c409ad77d62d186f1abdd/setup.py#L20 will change to tech-team@elifesciences.org

src/__init__.py_base
seanwiseman: Happy to change the format or version, but it does need a value to pin to in https://github.com/elifesciences/api-raml/blob/b05910d25fa34df276d10d81f3c3d12bef09739f/setup.py#L7 . Is '0.0.1' OK for now?

setup.py
seanwiseman: After testing `py@elifesciences.org` I got a delivery failure notification, so I think its safe to assume its no longer working.

It has now been changed:
https://github.com/elifesciences/api-raml/blob/df57ae9e9db08516ac73ec44d1d9509f15f4b820/setup.py#L13

thewilkybarkid (approved):

thewilkybarkid:
@giorgiosironi Think we missed a `Jenkinsfile` update to also commit `elife_api`.

thewilkybarkid:
Ah, was mentioned in https://github.com/elifesciences/api-raml/pull/195#discussion_r145141529.

PR #200, opened 2017-10-20
added dates received and accepted to article body

lsh-0:


thewilkybarkid (changes requested):
These are already in the history API - https://github.com/elifesciences/api-raml/blob/develop/src/model/article-history.v1.yaml#L5-L10

thewilkybarkid:
Just reread the ticket, moving back to that.

PR #203, opened 2017-11-09
Use newer block types in job adverts

thewilkybarkid:
Due to the merge timings in #178 and #186, job adverts were not updated for the new block types. The API is still marked as experimental, so it's ok to make this tiny BC break on something that probably won't happen (an advert isn't going to have a figure that has source data).


giorgiosironi (approved):
This is exactly one of the unknown unknowns that @experimental deals with

PR #204, opened 2017-11-28
ELPP-3304 Add access level to profile details

thewilkybarkid:
Breaking change, but it's an experimental API.

### DIFF
--- a/src/samples/profile/v1/complete.json
+++ b/src/samples/profile/v1/complete.json
@@ -11,7 +11,7 @@
             "access": "public"
         },
         {
-            "value": "j.carberry2@orcid.org",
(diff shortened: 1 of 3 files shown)


src/samples/profile/v1/complete.json
```diff
@@ -6,52 +6,65 @@
     },
     "orcid": "0000-0002-1825-0097",
     "emailAddresses": [
-        "j.carberry@orcid.org"
+        {
+            "value": "j.carberry@orcid.org",
+            "access": "public"
+        },
+        {
+            "value": "j.carberry2@orcid.org",
```
giorgiosironi: `j.carberry@restricted.example.com` or something that makes it clear it's private?

PR #206, opened 2017-12-21
Add filter for access level on annotations list

nlisgo:


src/api.raml
```diff
@@ -219,6 +219,16 @@ traits:
...
+            access:
+                description: |
+                    Return annotations appropriate for this access level.
+
+                    `restricted` will also return `public` annotations.
+                type: string
+                enum:
+                  - public
+                  - restricted
+                default: public
```
giorgiosironi, line 231: The default is `public`, which means until journal passes this parameter explicitly, it will only see public annotations.

Since journal's API consumer is in the `view-restricted-annotations` group, it is up to journal to pass this parameter with a correct value; for example, hiding private annotations from the profile of A when it is B seeing it, but allowing private annotations when it is A itself seeing the page.

giorgiosironi (approved):

thewilkybarkid (changes requested):
We need to say something like `if you have access`.

(We don't currently define `securitySchemes`.)

nlisgo:
This might be easier to agree on quickly when we are in the same location. I will bring it up next week.

thewilkybarkid (approved):

PR #207, opened 2017-12-22
There must be a value for highlight or content

nlisgo:


src/model/annotation-list.v1.yaml
```diff
@@ -54,12 +54,18 @@ properties:
                           - $ref: ../blocks/paragraph.v1.yaml
                           - $ref: ../blocks/quote.v1.yaml
                           - $ref: ../blocks/youtube.v1.yaml
+                    minItems: 1
```
nlisgo, line 57: this is needed to ensure the `content` in not empty.

src/samples/annotation-list/v1/first-page.json
```diff
@@ -50,6 +50,33 @@
...
+                "uri": "https://elifesciences.org/articles/14107"
+            },
+            "parents": [],
+            "created": "2017-09-15T16:40:00Z",
+            "content": [
+                {
+                    "type": "paragraph",
+                    "text": "Paragraph text"
+                }
+            ]
```
nlisgo, line 79: It would be great if we could have failing examples in this project but locally I did add an item with not `highlight` or `content` and it threw an error when I tried to compile it.

thewilkybarkid: Yeah, always meant to introduce a larger test suite with negative tests.

thewilkybarkid (approved):

PR #209, opened 2018-01-04
Let annotation parents be optional

thewilkybarkid:
Since some annotations can never have a parent, `parents` should be optional.


giorgiosironi:
Originally we had the choice of having an optional field, or an empty array; no particular preference and it ended up being the latter as it avoids clients having to check. If this is more in line with the style of other APIs, :+1:

nlisgo (approved):

PR #210, opened 2018-01-05
Add support for code in annotations

nlisgo:


thewilkybarkid (approved):
Looks to be possible through the Hypothesis interface, even if there's no button for it.

(Note, so are headings/sections...)

PR #211, opened 2018-01-23
Rename annotation parents to ancestors

thewilkybarkid:
Following https://github.com/elifesciences/journal/pull/837#discussion_r163224077.

Annotations will need to be updated to supply both while clients (ie Journal) switch over.

### DIFF
--- a/src/model/annotation-list.v1.yaml
+++ b/src/model/annotation-list.v1.yaml
@@ -18,7 +18,7 @@ properties:
                 access:
                     $ref: ../misc/access-level.v1.yaml
                 ancestors:
-                    description: IDs of previous annotations in this thread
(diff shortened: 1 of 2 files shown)


giorgiosironi (approved):
Given these are optional fields the change is small

src/model/annotation-list.v1.yaml
```diff
@@ -17,8 +17,8 @@ properties:
                     type: string
                 access:
                     $ref: ../misc/access-level.v1.yaml
-                parents:
-                    description: Parent annotation IDs
+                ancestors:
+                    description: IDs of previous annotations in this thread
```
giorgiosironi: order counts, we may state what comes first (oldest one to newest one?)

nlisgo (approved):

PR #212, opened 2018-01-24
observer, wrapping non-rest api endpoints

lsh-0:
rough attempt to capture the common attributes between all reports

### DIFF
--- a/src/api.raml
+++ b/src/api.raml
@@ -1495,6 +1495,33 @@ traits:
                                     Demonstration of the list being empty (the first page will always return a valid result).
                                 value: !include samples/recommendations/v1/empty.json
 
+/report/{name}:
+    type: collection
+    uriParameters:
+        name:
+            type: string
+            enum:
+                - latest-articles
+                - latest-articles-by-subject
+                - upcoming-articles
+                - published-research-article-index
+    get:
+        queryParameters:
+            format:
+                description: report serialisation type
+                type: string
+                required: false
+                example: csv
+        responses:
+            200:
+                body:
+                    text/csv:
+                    text/xml:
+                    application/json:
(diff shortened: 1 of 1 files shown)


src/api.raml
```diff
@@ -1495,6 +1495,33 @@ traits:
...
+        name:
+            type: string
+            enum:
+                - latest-articles
+                - latest-articles-by-subject
+                - upcoming-articles
+                - published-research-article-index
+    get:
+        queryParameters:
+            format:
```
giorgiosironi, line 1510: format is usually negotiated through the `Accept` header in the rest of the APIs, we may want to expose that for consistency (even if `observer` could still be implementing additional  parameters)

lsh-0: observer isn't REST-ful. the intent of observer is to make reports easy to consume by regular people. I don't see REST as user friendly. developer friendly, perhaps. Specifying formats through the `Accept` header isn't user friendly.

I'm looking for show stoppers that would prevent observer from being integrated into the api that absolutely must be added to observer before integration can happen.

this isn't to say REST-ful features can't be added later to make it more consistent, but we have a useful feature ready to be used by the public by an elife program that doesn't really have an API and it can get tacit support through the gateway or it can live outside of it. gateway support brings caching and url obfuscation. 

if you think format negotiation via the Accept header is a basic necessity here, I'll add it.

giorgiosironi: I don't think it's a blocker, just a normalization that can be added later. From the point of view of a client that is accessing several APIs it would ease the task of accessing this one; but given the format change there is always going to be some variety.

lsh-0:
Observer is too dynamic to properly capture in this document, closing.

PR #214, opened 2018-02-19
ELPP-3380 Add aims and scope to subjects

nlisgo:


thewilkybarkid (changes requested):
It's not a breaking change to add a new property so we don't need a new version. `content` also isn't the right term, content of a subject to me is the articles it covers ('Aims and scope' is the business term).

nlisgo:
Have made the changes. Should I drop the and from the property id? `aimsScope` or `aimsAndScope`?

src/model/subject.v1.yaml
```diff
@@ -7,6 +7,12 @@ allOf:
         impactStatement:
             description: Description of the subject
             type: string
+        aimsAndScope:
+            title: Aims and scope
+            type: array
+            items:
+                oneOf:
```
thewilkybarkid: Just use a plain `$ref`.

src/model/subject.v1.yaml
```diff
@@ -19,4 +25,5 @@ allOf:
               - thumbnail
     required:
       - impactStatement
+      - aimsAndScope
```
thewilkybarkid: The property also needs a `minItems` if it's `required`. But having it required means that we'll need data being present before this can be merged. So, let's make it optional for now and can make it required in the future (which also isn't a breaking change).

thewilkybarkid (approved):

src/api.raml
```diff
@@ -297,9 +297,6 @@ traits:
                         application/vnd.elife.annual-report+json;version=1:
                             schema: !include ../dist/model/annual-report.v1.json
                             examples:
-                                minimum:
-                                    displayName: Minimum
-                                    value: !include samples/annual-report/v1/minimum.json
```
thewilkybarkid: This needs to be reverted.

PR #215, opened 2018-03-14
ELPP-3324 Add article XML property

thewilkybarkid:
I've left it as optional for now, we could make it required once the data is populated.

### DIFF
--- a/src/samples/article-poa/v2/complete.json
+++ b/src/samples/article-poa/v2/complete.json
@@ -120,6 +120,7 @@
         }
     ],
     "pdf": "https://elifesciences.org/content/5/e14107.pdf",
+    "xml": "https://elifesciences.org/content/5/e14107.xml",
     "subjects": [
         {
             "id": "biochemistry",
--- a/src/misc/article.v2.yaml
+++ b/src/misc/article.v2.yaml
@@ -4,6 +4,9 @@ type: object
 allOf:
   - $ref: ../snippets/article.v1.yaml
   - properties:
+        xml:
+            type: string
+            format: uri
         issue:
             type: integer
             minimum: 1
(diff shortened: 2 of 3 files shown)


src/samples/article-poa/v2/complete.json
```diff
@@ -120,6 +120,7 @@
         }
     ],
     "pdf": "https://elifesciences.org/content/5/e14107.pdf",
```
giorgiosironi, line 122: can these samples be modernized with journal-style URLs?

giorgiosironi (approved):

src/samples/article-poa/v2/complete.json
thewilkybarkid: They should be buckets if they're to be 'real'.

giorgiosironi: Could make them clearly fake rather than fake but redirecting to the wrong page? (https://elifesciences.org/content/5/e14107.pdf goes to https://elifesciences.org/articles/14107)

gnott: In Crossref deposits, the XML and PDF URIs are set using the format in a configuration file (https://github.com/elifesciences/elife-crossref-xml-generation/blob/develop/crossref.cfg#L36-L37) loading the version's files from the CDN. If that is useful in the sample data at all.

I guess the value can be populated by ``bot-lax-adaptor`` when the property is available in the schema.

giorgiosironi: It should be, it's a predictable URL akin to the PDF

thewilkybarkid: Will leave the sample change for a separate PR.

PR #217, opened 2018-05-03
ELPP-3196 Add more types to metrics

thewilkybarkid:
Add more type options to metrics (not every type possible, just those that we need).

We don't need it split by time for these types, but rather than create a separate API we can use the existing one and just return the total in the response if it's simpler to implement. A `metric=downloads` request for these types should also 404.

### DIFF
--- a/src/api.raml
+++ b/src/api.raml
@@ -1163,6 +1163,11 @@ traits:
             type: string
             enum:
               - article
+              - blog-article
+              - event
+              - interview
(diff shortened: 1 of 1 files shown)


giorgiosironi (approved):
Seems to fit the existing types

lsh-0 (approved):
approving, but reserving the right to bitch and moan later :japanese_ogre:

PR #218, opened 2018-05-21
Reduce download size

thewilkybarkid:
Cut outs things like `src` from download files, which will mean smaller containers elsewhere.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,4 @@
 /*.* export-ignore
 /adr export-ignore
-/Jenkinsfile export-ignore
+/Jenkinsfile* export-ignore
 /src export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+/*.* export-ignore
+/adr export-ignore
+/Jenkinsfile export-ignore
+/src export-ignore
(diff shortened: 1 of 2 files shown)


.gitattributes
```diff
@@ -0,0 +1,4 @@
+/*.* export-ignore
```
thewilkybarkid, line 1: Leaves `LICENSE` in the root only.

.gitignore
```diff
@@ -1,4 +1,3 @@
 /composer.lock
-/dist/
```
thewilkybarkid, line 2: Will be ignored in archives otherwise.

giorgiosironi: Possible ramifications:

- https://github.com/elifesciences/api-raml/blob/develop/Jenkinsfile#L10 should not be affected
- https://github.com/elifesciences/api-raml/blob/develop/Jenkinsfile#L19 uses [`git add --all --force`](https://github.com/elifesciences/elife-jenkins-workflow-libs/blob/master/vars/elifeGitAutoCommit.groovy#L7) so should also not be affected

giorgiosironi (approved):

.gitignore
thewilkybarkid: RAMLifications.

thewilkybarkid:
Made 0.5 MB smaller.

PR #219, opened 2018-06-27
ELPP-3738 Add new article type 'research-communication'

lsh-0:
ticket: https://elifesciences.atlassian.net/browse/ELPP-3738

### DIFF
--- a/src/api.raml
+++ b/src/api.raml
@@ -1551,6 +1551,7 @@ traits:
                       - insight
                       - research-advance
                       - research-article
+                      - research-communication
                       - retraction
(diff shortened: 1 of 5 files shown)


thewilkybarkid:
There's more changes than this, I'll add them.

Technically this is a BC break, but let's see the extend of the changes downstream (suspect we'll need to be pragmatic and not create new versions).

lsh-0:
> There's more changes than this, I'll add them.

knew it couldn't be that easy ;)

and I haven't considered the downstream effects any farther than lax.

PR #220, opened 2018-07-04
Fix IIIF identifiers

thewilkybarkid:


src/misc/image.v1.yaml
```diff
@@ -8,7 +8,7 @@ allOf:
             title: IIIF Image
             type: string
             format: uri
-            pattern: ^https://
+            pattern: ^https://.+/[^/?#\[\]@]+$ # http://iiif.io/api/image/2.1/#uri-encoding-and-decoding, with support for percent encoding
```
thewilkybarkid: Should improve the possible prefix to have percent-encode special characters (but allow slashes).

giorgiosironi: Seems `https://iiif.elifesciences.org/lax:09560/elife-09560-fig1-v1.tif/full/full/0/default.jpg` has a lot of slashes? `https://iiif.elifesciences.org/lax:09560%2Felife-09560-fig1-v1.tif` is the example that should match this?

thewilkybarkid: Yep. (This property would have been better as pointing to the `info.json` file, would have made it less ambigious. This is the base URI: `{scheme}://{server}{/prefix}/{identifier}`)

src/misc/image.v1.yaml
```diff
@@ -8,7 +8,7 @@ allOf:
             title: IIIF Image
             type: string
             format: uri
-            pattern: ^https://
+            pattern: ^https://[^/]+(/[^?#\[\]@]+)?/[^/?#\[\]@]+$ # http://iiif.io/api/image/2.1/#uri-encoding-and-decoding, with support for percent encoding
```
thewilkybarkid: Doesn't test for a valid authority. Assuming `format: uri` covers that.

giorgiosironi: The first subgroup looks like an optional prefix like the IIIF server location, but then it seems to match the old URLs like `https://iiif.elifesciences.org/lax:09560/elife-09560-fig1-v1.tif` https://regex101.com/r/NngB53/3

thewilkybarkid: Yep, `lax:09560` is a valid prefix and `elife-09560-fig1-v1.tif` a valid identifier. (This is why slashes in identifiers needs to be escaped.)

giorgiosironi: (also `https://iiif.elifesciences.org/lax/09560/elife-09560-fig1-v1.tif` as the end of the IIIF server path is indistuinguishable from the start of the IIIF file path)

src/misc/image.v1.yaml
```diff
@@ -8,7 +8,7 @@ allOf:
             title: IIIF Image
             type: string
             format: uri
-            pattern: ^https://[^/]+(/[^?#\[\]@]+)?/[^/?#\[\]@]+$ # http://iiif.io/api/image/2.1/#uri-encoding-and-decoding, with support for percent encoding
+            pattern: ^https://[^/]+(?:/[^?#\[\]@]+)?/[^/?#\[\]@]+$ # http://iiif.io/api/image/2.1/#uri-encoding-and-decoding, with support for percent encoding
```
giorgiosironi: (previous comments vanished on the outdated diff)

giorgiosironi: to make it strict enough, it may have to mandate no IIIF server prefix hence no `/` after `https://hostname/`

thewilkybarkid: We should switch to using proper prefixes rather than Loris's manipulation of the identifier (eg `https://iiif.elifesciences.org/lax/09560%2Felife-09560-fig1-v1.tif`).

thewilkybarkid: Happy to :shipit:?

giorgiosironi (approved):
Can't think of a stricter pattern

PR #221, opened 2018-07-09
Add digests API

thewilkybarkid:
/cc @gnott @nlisgo


gnott:
I thought the image for a digest may be optional, but I may have that wrong. 

When there is an image, it may have a caption and an attribution, which I think the schema may support, even though ``image`` may not technical define a caption as a property.

I'm not sure how the related article snippet details will be added, not that it cannot be added, I'm just unsure of exactly how it will go together.

thewilkybarkid:
> I thought the image for a digest may be optional, but I may have that wrong.

My understanding is that it's required (we're going to run into design problems if there isn't one).

> When there is an image, it may have a caption and an attribution, which I think the schema may support, even though image may not technical define a caption as a property.

The image will need to be in two places, `thumbnail` and in an `image` block inside `content` (which can then have a caption, attribution etc).

> I'm not sure how the related article snippet details will be added, not that it cannot be added, I'm just unsure of exactly how it will go together.

A starting point would be to follow Recommendations and fetch it at runtime. This isn't great as it takes longer (ie have to make a HTTP request), but is straightforward to do. A better approach is to store the details locally, but this involves caching busting etc to keep it up to date.

gnott:
One more question if this is a good place to ask: will the ``id`` attribute of the digest be set as the article's ``msid`` that the digest is associated with? This should always be unique. Or, will it be assigned by some underlying data store and does not relate to the content itself?

thewilkybarkid:
The ID just needs to be a unique string, doesn't matter what it is. As a digest can be linked to more than one article, might be better to just generate a value (like we do for blog posts etc). Edit: eg https://medium.com/life-on-earth/the-continuing-mysteries-of-homo-naledi-53836178c687

thewilkybarkid:
Going to roll back the changes for the other APIs (collections, highlights and search) so that the digests API can be merged first.

src/model/digest.v1.yaml
```diff
@@ -0,0 +1,24 @@
+$schema: http://json-schema.org/draft-04/schema#
+title: Digest
+type: object
+allOf:
+  - $ref: ../snippets/digest.v1.yaml
+  - properties:
+        content:
```
giorgiosironi, line 7: https://github.com/elifesciences/api-validator-python will come in handy for whatever implements this

src/snippets/digest.v1.yaml
```diff
@@ -0,0 +1,36 @@
+$schema: http://json-schema.org/draft-04/schema#
+title: Digest snippet
+type: object
+properties:
+    id:
+        $ref: ../misc/id.v1.yaml
+    title:
+        description: Title
+        type: string
```
giorgiosironi, line 9: do we ned to say "not empty"?

giorgiosironi (approved):

src/model/digest.v1.yaml
gnott: I didn't know this library provided the ability. I am playing with implementing some JSON output in the https://github.com/elifesciences/digest-parser project. Python 3 would be ok there, but if we want the elife-bot to run it, then the min Python 3.5 of api-validator-python might need some creative solutions. Maybe dockerise the conversion or something?

src/samples/digest/v1/minimum.json
```diff
@@ -0,0 +1,42 @@
+{
+    "id": "1",
```
gnott, line 2: For the ``id`` value, my suggestion is something like ``"digest-99999"``, the ``99999`` being the article the digest is associated with. There are no constraints for me on this value, so please suggest an alternate if you don't like it.

gnott (approved):

src/samples/digest/v1/minimum.json
thewilkybarkid: I wouldn't include the type in the ID, and would be wary of using an article ID as it's a `1..n` relationship not `1..1`. But that's not important for now as it's just a string.

PR #225, opened 2018-07-24
Add optional caption to youtube block

nlisgo:
See: https://github.com/elifesciences/jira-import/issues/3555


src/blocks/youtube.v1.yaml
```diff
@@ -1,22 +1,24 @@
...
-        type: integer
-        minimum: 1
+allOf:
+  - $ref: ../misc/caption.v1.yaml
+  - properties:
+        type:
+            type: string
+            enum:
+              - youtube
+        id:
```
thewilkybarkid, line 11: This now conflicts with the `id` property inside `caption.v1.yaml` (which is meant for the HTML ID).

I'm wondering whether we should forget YouTube as a specific type, and just use an `embed` block type instead (which allows for arbitrary things to be used). That's a more complicated change though.

thewilkybarkid:
Are there tickets for these changes?

nlisgo:
> Are there tickets for these changes?

I've referenced the issue in the description.

src/blocks/youtube.v1.yaml
```diff
@@ -17,6 +17,12 @@ properties:
         title: Height
         type: integer
         minimum: 1
+    caption:
+        title: Caption
+        type: array
+        items:
+            oneOf:
+              - $ref: ../blocks/paragraph.v1.yaml
```
nlisgo: @thewilkybarkid since we only need the `caption` property, could this be the best solution?

thewilkybarkid: On the right lines! `title` and `caption` are a bit tied together in various places, but I don't _think_ `id` is, so we should maybe include in separately where needed (eg `video.v1.yaml`). Not a breaking change then.

src/blocks/youtube.v1.yaml
nlisgo: This is no longer a conflict since the `id` property has been removed from `caption.v1.yaml` and must be referenced separately.

src/blocks/youtube.v1.yaml
```diff
@@ -2,23 +2,30 @@ $schema: http://json-schema.org/draft-04/schema#
 title: YouTube video
 type: object
 properties:
```
thewilkybarkid: This needs to be `allOf`.

src/blocks/youtube.v1.yaml
```diff
@@ -2,23 +2,30 @@ $schema: http://json-schema.org/draft-04/schema#
...
+            type: string
+        width:
+            title: Width
+            type: integer
+            minimum: 1
+        height:
+            title: Height
+            type: integer
+            minimum: 1
+        title:
```
thewilkybarkid: Not needed as is part of `caption.v1.yaml`.

src/samples/labs-post/v2/complete.json
```diff
@@ -48,7 +48,8 @@
             "type": "youtube",
             "id": "-9JVFCL0fvk",
             "width": 960,
-            "height": 720
+            "height": 720,
+            "title": "A sneak peek at Manuscripts, a new writing tool that helps you write an insanely great paper."
```
thewilkybarkid: We should add a `caption` too (even if we're not going to use it for now).

thewilkybarkid (approved):
Couple of tweaks, looks good apart from that.

Worth doing the downstream work before merging though just to make sure there's no gotchas.

src/blocks/youtube.v1.yaml
```diff
@@ -1,9 +1,8 @@
 $schema: http://json-schema.org/draft-04/schema#
 title: YouTube video
 type: object
-properties:
+allOf:
   - $ref: ../misc/caption.v1.yaml
-  - $ref: ../misc/attribution.v1.yaml
```
nlisgo: This was added in error. There has been no request for attribution for youtube videos.

nlisgo:
I am now working on the downstream work. Will reference the PR's here.

giorgiosironi (approved):
The changes make sense, although the slight overlap of `id`-related types that is however expressed in the schema. Fields are optional and hence they are perfectly backward compatible.

PR #228, opened 2018-08-10
Digests write side

giorgiosironi:
For https://github.com/elifesciences/issues/issues/4374

Added what I can for the POST, to check everything compiles. Some things lifted from https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md


src/api.raml
```diff
@@ -832,6 +832,44 @@ traits:
...
+                                displayName: Minimum
+                                value: !include samples/digest/v1/minimum.json
+                            complete:
+                                displayName: Complete
+                                value: !include samples/digest/v1/complete.json
+            400:
+                description: Missing or invalid fields for digest.
+            403:
+                description: Not authorized to add a digest.
+                #body:
```
giorgiosironi: would add, but not implemented on 403?

src/api.raml
```diff
@@ -832,6 +832,44 @@ traits:
                                 description: |
                                     Demonstration of the list being empty (the first page will always return a valid result).
                                 value: !include samples/digest-list/v1/empty.json
+    post:
+        description: |
+            Add a new digest.
```
giorgiosironi: could add request headers for authorization, but can either be `Authorization` (for the `api-gateway`) or `X-Consumer-Group` (probably wrong, for the backend service)

src/api.raml
thewilkybarkid: Need to include type, schema and example like https://github.com/elifesciences/api-raml/blob/534c9dde2965337a91f5a41b0b9744d644e87249/src/api.raml#L84-L103.

src/api.raml
```diff
@@ -832,6 +832,44 @@ traits:
...
+                    minimum:
+                        displayName: Minimum
+                        value: !include samples/digest/v1/minimum.json
+                    complete:
+                        displayName: Complete
+                        value: !include samples/digest/v1/complete.json
+        responses:
+            200:
+                description: |
+                    Digest just created.
```
thewilkybarkid: `201`, not `200`.

src/api.raml
```diff
@@ -832,6 +832,44 @@ traits:
                                 description: |
                                     Demonstration of the list being empty (the first page will always return a valid result).
                                 value: !include samples/digest-list/v1/empty.json
+    post:
```
thewilkybarkid: Given the ID is coming from outside, I'd rather just have `PUT`.

giorgiosironi: I think the ID can be auto-generated, although coming it from outside is safer for idempotence. We could document here the PUT and PATCH and deprecate the POST.

src/api.raml
```diff
@@ -837,6 +847,42 @@ traits:
                                 complete:
                                     displayName: Complete
                                     value: !include samples/digest/v1/complete.json
+        put:
+            description: |
+                Upsert a digest.
```
gnott, line 852: Where in the dictionary will I find `"upsert"`? :wink:

gnott (approved):
`put` looks good, simple and includes the status codes for success and failure :+1:

src/api.raml
```diff
@@ -837,6 +847,42 @@ traits:
                                 complete:
                                     displayName: Complete
                                     value: !include samples/digest/v1/complete.json
+        put:
```
giorgiosironi, line 850: I think we are dropping `POST` (id is always passed in through `PUT`) and `PATCH` (incorrectly implemented, difficult to do) so this verb should suffice

giorgiosironi (commented):
Cannot approve because I opened the PR, but :+1: on the latest version

src/api.raml
giorgiosironi: `(verb, adjective, noun): to cause discomfort to a data storage entity e.g. the database table was upsert that the INSERT query hurt her feelings`

PR #229, opened 2018-08-10
Add decision letter and author response IDs

thewilkybarkid:


thewilkybarkid:
I've left it optional for now. We can make it required in the future when the data has been backfilled (though there's no rush on that).

PR #230, opened 2018-08-10
Drop annual-report and annual-report-list version 1

nlisgo:
Improves: elifesciences/issues#4391


src/api.raml
```diff
@@ -260,21 +260,7 @@ traits:
...
-                                description: |
-                                    List is split over two pages.
-                                value: !include samples/annual-report-list/v1/populated.json
-                            empty:
-                                displayName: Empty list
-                                description: |
-                                    Demonstration of the list being empty (the first page will always return a valid result).
-                                value: !include samples/annual-report-list/v1/empty.json
                     application/vnd.elife.annual-report-list+json;version=2:
-                        (experimental):
```
thewilkybarkid, line 277: If anyone else was using this we should mark the other as deprecated prior to removing. But they're not.

thewilkybarkid (approved):

PR #231, opened 2018-09-04
Modify /people for xPub/eJP integration requirements

diversemix:
As part of the requirements for the xPub/eJP integration we need to have given names, surname and email. This will allow eJP to reconcile a user within their system when importing a manuscript.

The field names were selected from JATS, see:

image


src/misc/person.v1.yaml
```diff
@@ -26,12 +26,37 @@ properties:
...
+                title: Surname
+                description: |
+                    The surname of the person.
+
+                    For example:
+                      - Schekman
         required:
           - preferred
           - index
+          - last
```
BlueReZZ: Should this be `surname` now?

BlueReZZ: Also, if making this new field required means backward compatibility issues - is that the right step forward?

nlisgo: We don't need this to be required for all people, only those that xPub is pulling in. So, it should remain optional for now. We could make it required and bump the version if we can populate the last name for all people. Leave as optional for now.

src/misc/person.v1.yaml
```diff
@@ -26,12 +26,37 @@ properties:
...
+        type: object
+        properties:
+            email:
+                title: Email address
+                description: This person's email for correspondence.
+            required:
+                - email
 required:
   - name
+  - correspondence
```
nlisgo: This is not a candidate for being required as most of the BRE we have no need to store their email addresses.

nlisgo (changes requested):

src/misc/person.v1.yaml
thewilkybarkid: It's not a breaking change to add a new required field to a response, but does require the data being available everywhere.

src/misc/person.v1.yaml
```diff
@@ -26,12 +26,37 @@ properties:
...
+                      - Schekman
         required:
           - preferred
           - index
+          - last
     orcid:
         title: ORCID
         type: string
         pattern: ^[0-9]{4}-[0-9]{4}-[0-9]{4}-[0-9]{3}[0-9X]$
+    correspondence:
```
thewilkybarkid: We should make this consistent with https://github.com/elifesciences/api-raml/blob/c3aaefd514b991848c3254cd7a4633c4e64e96b2/src/misc/author.v1.yaml#L36-L40

src/misc/person.v1.yaml
```diff
@@ -26,12 +26,37 @@ properties:
                       - Schekman, Randy
                       - VijayRaghavan, Krishnaswamy
                       - Li Wenhui
+            given_names:
```
thewilkybarkid: We've used lower camel case elsewhere.

src/misc/person.v1.yaml
```diff
@@ -26,12 +26,37 @@ properties:
...
+                title: Given names
+                description: |
+                    The given names of the person.
+
+                    For example:
+                      - Randy
+            surname:
```
thewilkybarkid: We originally went for just single strings as names are a _very_ thorny issue.

(Caveat to the following: might be completely wrong.) I believe one of our senior editors has a name in the format `[Father's name] [Given name]`, which looks to get rather mangled (including in our own JATS).

Unfortunately we're dealing with imperfect inputs and outputs. I'm tempted to say we should try and avoid componentising it here (since we can't do it properly) and instead just have 'first names' and 'last names' (with the latter optional). Or a list of 1 or 2 values. Makes it ambiguous as to what these names are, but that's the point.

diversemix: Yeah. This has been an ongoing issue with eJP, apparently they need to have the `surname`  to reconcile users, and they use the JATS one. The idea is that we don't replace the existing `index` and `preferred` but complement with fields for given name and surname ( that staff can fill out in Journal)

diversemix: @thewilkybarkid ... are you happy if I resolve this conversation?

thewilkybarkid: I'll never be happy with it, but yes. 😉

diversemix:
@nlisgo you happy with this now?

nlisgo (approved):

PR #232, opened 2018-09-25
adds 'digests' to list of content types supported by elife-metrics

lsh-0:


thewilkybarkid (approved):

src/api.raml
```diff
@@ -1270,6 +1270,7 @@ traits:
               - article
               - blog-article
               - collection
+              - digests
```
thewilkybarkid: Hang on, needs to be singular!

thewilkybarkid (changes requested):

src/api.raml
lsh-0: I had completely overlooked that. I'll double check the corresponding changes in my elife-metrics PR

thewilkybarkid (approved):

PR #233, opened 2018-09-26
Support digests in highlight lists and collections

nlisgo:
Improves: https://github.com/elifesciences/issues/issues/4294


thewilkybarkid (approved):

thewilkybarkid:
Just spotted https://github.com/elifesciences/issues/issues/4379, since we're doing a version bump should do this at the same time.

nlisgo:
Have added support for events in collections. I am assuming that events should be displayed in a collection even after the event has taken place.

thewilkybarkid (approved):

PR #234, opened 2018-11-27
support for PUT and POST methods on article version endpoint

lsh-0:
I removed the poa/vor version=1 responses just for brevity (but they are deprecated right? it would be nice to remove them at some point).

### DIFF
--- a/src/api.raml
+++ b/src/api.raml
@@ -506,29 +506,19 @@ traits:
                             Article version number.
                         type: integer
                         minimum: 1
-                get:
-                    description: |
-                        Get an article.
+                put:
+                    description: Ingest an article.
+                    uriParameters:
+                        force:
+                            description: Force the operation
+                            type: boolean
+                        dry-run:
+                            description: Revert operation
+                            type: boolean
                     responses:
                         200:
-                            description: |
-                                Article version.
+                            description: Article version.
                             body:
-                                application/vnd.elife.article-poa+json;version=1:
-                                    schema: !include ../dist/model/article-poa.v1.json
-                                    examples:
-                                        minimum:
-                                            displayName: Minimum
-                                            value: !include samples/article-poa/v1/minimum.json
-                                        complete:
-                                            displayName: Complete
-                                            value: !include samples/article-poa/v1/complete.json
-                                        unpublishedMinimum:
-                                            displayName: PreviewMinimum
-                                            value: !include samples/article-poa/v1/unpublished-minimum.json
-                                        unpublishedComplete:
-                                            displayName: PreviewComplete
-                                            value: !include samples/article-poa/v1/unpublished-complete.json
                                 application/vnd.elife.article-poa+json;version=2:
                                     (experimental):
                                     schema: !include ../dist/model/article-poa.v2.json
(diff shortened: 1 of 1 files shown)


lsh-0:
A better approach than a POST body with a token might be putting the token into a header.

src/api.raml
```diff
@@ -506,29 +506,65 @@ traits:
                             Article version number.
                         type: integer
                         minimum: 1
-                get:
-                    description: |
-                        Get an article.
+                put:
+                    description: Ingest an article.
+                    queryParameters:
```
giorgiosironi, line 511: never seen query string parameters on a `PUT`; analyzing what they do:

`force` allows the `PUT` to have different results depending on the existing state, probably silent correction vs new ingest?
`dry-run` doesn't change the state but only validates everything will be fine when not used.

Both use cases seem to be served by different resources?
- `/article/:id/versions/:v:/`
- `/article/:id/versions/:v:/dry-run`
but it would be helpful to elaborate on their intended meaning to decide, especially `force`

src/api.raml
```diff
@@ -506,29 +506,65 @@ traits:
                             Article version number.
                         type: integer
                         minimum: 1
-                get:
-                    description: |
-                        Get an article.
+                put:
```
giorgiosironi, line 509: the `put` doesn't seem to define the body of the request (e.g. from `digests` https://github.com/elifesciences/api-raml/blob/a1896a457b4f0a22d314293706bf6aa6aecae04f/src/api.raml#L927), but only of the response.

Response may be omitted if not needed, but it seems good to have it especially if there is additional logic like adding dates from the server side.

src/api.raml
```diff
@@ -506,29 +506,65 @@ traits:
...
+                                        complete:
+                                            displayName: Complete
+                                            value: !include samples/article-vor/v2/complete.json
+                                        unpublishedMinimum:
+                                            displayName: PreviewMinimum
+                                            value: !include samples/article-vor/v2/unpublished-minimum.json
+                                        unpublishedComplete:
+                                            displayName: PreviewComplete
+                                            value: !include samples/article-vor/v2/unpublished-complete.json
+                post:
```
giorgiosironi, line 555: also the `post` doesn't define the body of the request, which should probably include the `queryParameters` instead (needs similar elaboration as the `put` on use cases in which these parameters are used)

giorgiosironi:
> A better approach than a POST body with a token might be putting the token into a header.

Which token? If you mean the `Authorization` header, https://github.com/elifesciences/api-raml/blob/b099f8c0aee6a6ac323570983c6e078cbad2236a/src/api.raml#L70-L78 and https://github.com/elifesciences/api-raml/blob/b099f8c0aee6a6ac323570983c6e078cbad2236a/src/api.raml#L863 may help.

src/api.raml
thewilkybarkid: `force` sounds like [conditional requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Conditional_requests)?

What does `dry-run` do that validating against the schema wouldn't?

src/api.raml
thewilkybarkid: `POST` on a specific version is odd. What does it mean?

src/api.raml
thewilkybarkid: I wouldn't expect a response body.

src/api.raml
lsh-0: > POST on a specific version is odd. What does it mean?

~creating a new~ *publishing* an article version

src/api.raml
lsh-0: > /article/:id/versions/:v:/dry-run

it would be `/article/:id/verisons/:v?dry-run=true`

> What does dry-run do that validating against the schema wouldn't?

1. lax contains business rules about what can be stored. it protects against stupid things, like trying to ingest or publish a version 2 when there is no version 1.

2. the article-json isn't complete coming out of bot-lax, there is information that is required that can't be known without knowing the complete article history. Validation occurs prior to lax with placeholders, dummy values that are then removed in lax.

3. lax also accumulates article 'fragments', pushed by journal-cms, and what is served up is the merged and validated result of all of these fragments.

4. bot-lax has an API that is used by Exeter to test articles are correct. the xml is transformed -> sent to lax -> validated and then the result is rolled back.

lsh-0:
> Which token? If you mean the Authorization header,

elife-bot sends a token through on it's requests that is used to identify responses (it communicates over SES/SQS).

giorgiosironi:
> elife-bot sends a token through on it's requests that is used to identify responses (it communicates over SES/SQS)

I think this should not necessarily be transmitted through this API, as it could be managed purely inside the `bot-lax-adaptor` (and being a bot-specific concern, it would be doing its job).

src/api.raml
giorgiosironi: 1 to 3 are all good concerns, whereas 4 seems to be the only scenario that involves calling the API with `dry-run`. That seems to me a different endpoint with a different implementation, e.g. `/article/:id/tentative-versions/:v`

lsh-0:
the idea of tracing a request throughout an app is still hanging around in my head, using the token to achieve part of that in lax where I could was still seems like a good idea. I don't know how to implement it properly though without completely hamming it up, so it might be a good idea to drop it.

The other concern I had was spurious POSTs accidentally publishing an article. a POST just seems like such an easy thing to do and get wrong that a speed bump doesn't seem like a bad idea.

lsh-0:
I dropped the HTTP ingest changes in lax years ago after zero interest. It might get resurrected one day. Closing.

PR #235, opened 2019-01-29
[WIP] Feature: adds 'publicationHistoryEvent' to articles

lsh-0:
Reverse link: https://github.com/elifesciences/issues/issues/4702

Modelling:

```
<pub-history>
    <event event-type="preprint-publication">
        <event-desc>This article was ogininally published as a <ext-link ext-link-type="uri" 
            xlink:href="https://www.biorxiv.org/content/early/2017/03/24/118356">preprint on Biorxiv</ext-link>
        </event-desc>
        <date iso-8601-date="2016-03-01">
            <day>24</day>
            <month>03</month>
            <year>2017</year>
        </date>  
    </event>
</pub-history> 
```


lsh-0:
I'm not clear what the distinction is between 'snippet', 'misc' and 'blocks' so I hope the bits are living in the right place.

I also assumed that the ./src/snippets/article.v1.yaml forms the basis of POA/VOR articles and is included in v2 as well.

src/snippets/article.v1.yaml
```diff
@@ -89,6 +89,10 @@ properties:
                 minItems: 1
         required:
           - content
+    publicationHistory:
```
giorgiosironi: the snippet is used in all listings so this seems a bit too much detail for that use case.
What about the schema of `/article/:id/versions`:
https://github.com/elifesciences/api-raml/blob/1524e9126fffdc16f2de7549d79666535d5dfe40/src/model/article-history.v1.yaml
or the general body of every `/article/:id` (for both POA and VOR): 
https://github.com/elifesciences/api-raml/blob/1524e9126fffdc16f2de7549d79666535d5dfe40/src/misc/article.v2.yaml

src/misc/publicationHistoryEvent.v1.yaml
```diff
@@ -0,0 +1,16 @@
...
+        type:
+            type: string
+            enum:
+                - preprint-publication
+        description:
+            type: string
+        externalLink:
+            type: string
+            format: uri
+        date:
```
giorgiosironi: there are names like `versionDate` and `statusDate` around, so this may be the generic name of those

src/misc/publicationHistoryEvent.v1.yaml
```diff
@@ -0,0 +1,16 @@
...
+title: Publication History Event
+type: object
+allOf:
+    - properties:
+        type:
+            type: string
+            enum:
+                - preprint-publication
+        description:
+            type: string
```
giorgiosironi: theoretically someone could put HTML in here, do we have anything stricter than `string`?

giorgiosironi: probably not: https://github.com/elifesciences/api-raml/blob/40b376afaa33083a9a86dd3516f171e091e7ffd0/src/model/subject.v1.yaml#L7-L9

src/snippets/article.v1.yaml
gnott: Looking at `<event>` (https://jats.nlm.nih.gov/archiving/tag-library/1.2d2/element/event.html) and its intent, it allows you to add some additional information about a publication event (or non-publication event). It seems to me more closely aligned with the `accepted` and `received` dates found currently in the `<history>` tag. These dates are associated with all versions of the article and, although the preprint date would not change between versions, it may not be precise to associate a preprint with article itself, and might be better aligned with the `article-history.v1.yaml` somehow? I don't have a concrete example because it is not easy.

src/misc/publicationHistoryEvent.v1.yaml
lsh-0: In the example Melissa gave us of it's intended usage (it's in the PR description), the description contains the external link embedded within the text. I think I argued against this, extracting the URI from an attribute or discrete element elsewhere would have been less ambiguous and the description may contain many such links. I don't think plain or escaped html would get through production, their standards are quite high.

lsh-0:
I've moved the definition inline with the current article history endpoint.

gnott:
I have a good feeling about the structure on this one, great job @lsh-0 !

src/model/article-history.v1.yaml
```diff
@@ -1,25 +1,50 @@
...
+          - accepted
+        accepted:
+          - received
+    required:
+        - versions
+-
+    title: Event
+    type: object
+    allOf:
+      - $ref: ../snippets/event.v1.yaml
```
giorgiosironi: but this is a [real life event](https://github.com/elifesciences/api-raml/blob/develop/src/samples/event/v2/complete.json), not a state change

lsh-0: a state change isn't a type of event? Are you objecting to my use of 'event' (which is fair enough, I didn't know that other event existed). I'm working with very little input here, if you have better ideas then please take a crack at it.

lsh-0: I could call it HistoryEvent? or HistoricalEvent??

gnott: I like `HistoryEvent`. You wouldn't need a `starts`, `ends` and `published` dates if you defined a new snippet, like the current `event` snippet requires.

lsh-0: oh - I see what I've done. This is a typo, that `- $ref: ../snippets/event.v1.yaml` isn't meant to be there.

lsh-0: actually it looks all kinds of messed up.

lsh-0: ok, just pushed what I intended to commit. it doesn't `node compile` so I'm investigating why

lsh-0: fixed. if this is more acceptable I'll write some samples

src/model/article-history.v1.yaml
```diff
@@ -1,25 +1,42 @@
...
+            minItems: 1
+    dependencies:
+        received:
+          - accepted
+        accepted:
+          - received
+    required:
+        - versions
+-
+    title: Publication Event
```
giorgiosironi, line 29: this is more in line with what I meant, not tying this to the other data type of "marketing" events

lsh-0:
May I get some more eyes on this please? There was a bad commit I had to fix up but it looks ok now.

gnott (approved):
Looks pretty clean and comprehensive to me 👍 Whether the description field is enough to display it correctly on Journal I'm not sure.

lsh-0:
Thanks @gnott 

@thewilkybarkid , if you have the time I'd like an opinion from you. My edits to the spec feel naive.

lsh-0:
I'll be working on this tomorrow, last opportunity to review before I merge

thewilkybarkid:
@lsh-0 Please don't merge this. This feature (ie getting it to appear in Journal) isn't currently in our to do list, and AFAIK we don't have a spec as to what the output would need to be anyhow.

/cc @Maelplaine @BlueReZZ

nlisgo:
@lsh-0 closing. Reopen or create an issue if you want to revisit this.

PR #236, opened 2019-03-12
Support press-package in highlight item

nlisgo:
Improves: https://github.com/elifesciences/issues/issues/4796


nlisgo:
Rowena on slack:

> Hi Nathan, press-package in a collection has not been a requirement historically, but that's not to say it wouldn't be in the future. If it's easy to add that functionality at the same time please do, that way if we need it in the future it's already.

src/model/highlight-list.v3.yaml
```diff
@@ -0,0 +1,91 @@
...
+                                              - podcast-episode
+                                  - $ref: ../snippets/podcast-episode.v1.yaml
+                              - $ref: ../misc/podcast-episode-chapter.v1.yaml
+                              - allOf:
+                                  - properties:
+                                        type:
+                                            type: string
+                                            enum:
+                                              - press-package
+                                  - $ref: ../snippets/press-package.v1.yaml
```
nlisgo, line 85: The only difference between `highlight-list.v2.yaml` and `highlight-list.v3.yaml` is the addition of the support for `press-package`.

See: https://github.com/elifesciences/api-raml/pull/236/files#diff-5b5d735f3567c09e9e0068fb93588f71R79-R85

nlisgo:
@thewilkybarkid, it's tempting to add support for press-package in collections at the same time as we did when we added digests to highlight lists: https://github.com/elifesciences/api-raml/pull/233

Thoughts? Or shall we just roll with this one?

thewilkybarkid:
Unless it saves time anywhere, probably wouldn't bother.

thewilkybarkid (approved):

PR #237, opened 2019-04-02
Add review-article type

gnott:
Re issue https://github.com/elifesciences/issues/issues/4766

These are some places I think `review-article` type might go. I found some of these just searching for other article types so the search API changes I just followed existing examples (I'm not sure of their further implications).


thewilkybarkid:
Like https://github.com/elifesciences/api-raml/pull/219#issuecomment-400557000 this is a BC break, but let's just manage it.

thewilkybarkid (approved):
Think that's all the cases, will merge in a bit.

giorgiosironi (approved):

PR #238, opened 2019-06-24
Feature: adds bioprotocol proposal

lsh-0:


src/model/bioprotocol.v1.yaml
```diff
@@ -0,0 +1,25 @@
...
+            type: string
+        protocol_title:
+            description: "section title, looks like 'Isolation of LDs and treatments'"
+            type: string
+        is_protocol:
+            type: boolean
+        protocol_status:
+            description: "numeric code corresponding to the current status of a protocol"
+            type: "integer"
+            minimum: 0
```
giorgiosironi: anything better than a number? Is there a translation table from `0`, `1` to keywords?

lsh-0: there isn't

src/model/bioprotocol.v1.yaml
```diff
@@ -0,0 +1,25 @@
...
+    type: object
+    properties:
+        protocol_sequencing_number:
+            description: "section ID, looks like 's4-1'"
+            type: string
+        protocol_title:
+            description: "section title, looks like 'Isolation of LDs and treatments'"
+            type: string
+        is_protocol:
+            type: boolean
```
giorgiosironi: always `false` in the example, what does it mean? An enumeration can be more extensible than a boolean e.g. `type`: `protocol`|`something-else` allows addition of other types.

src/model/bioprotocol.v1.yaml
```diff
@@ -0,0 +1,25 @@
+$schema: http://json-schema.org/draft-04/schema#
+title: BioProtocol
+type: array
+items:
+    type: object
+    properties:
+        protocol_sequencing_number:
+            description: "section ID, looks like 's4-1'"
+            type: string
+        protocol_title:
```
giorgiosironi: is the `protocol_` prefix in these two fields necessary? Thinking about e.g. [podcast episodes](https://github.com/elifesciences/api-raml/blob/develop/src/snippets/podcast-episode.v1.yaml#L5) where we use specific terms like `number` without a prefix.

src/model/bioprotocol.v1.yaml
lsh-0: I've asked Dong to elaborate

lsh-0: Dong has gotten back to us. They actually switched from a boolean *to* an integer but his reply doesn't read like they intend to expand the number of different states any time soon.

I can store the integer and return a boolean value instead. When/if BP decide to add more statuses we can tweak things then.

src/model/bioprotocol.v1.yaml
lsh-0: this is a flag they use when they determine that a protocol that has been given to them isn't actually a protocol - eLife mixes materials and methods together in the same section with no means of differentiation, and there is some talk at the moment on how to best deal with that.

```
$ curl https://dev.bio-protocol.org/api/elife00005 | jq '.Protocols[].IsProtocol'`
true
true
true
true
true
true
```

I'll clarify with them, but I think what I'll do is exclude rows from being returned to the journal where `IsProtocol=false` so that a link isn't generated. And I'll exclude this field as well as it will then be redundant.

lsh-0: I've pinged Dong about it. It seems the most obvious and reasonable thing to do and they seem quite competent so I'll go ahead and make the changes. I can always revert if they start going crazy.

src/model/bioprotocol.v1.yaml
lsh-0: I can remove those

src/model/bioprotocol.v1.yaml
lsh-0: yes, Dong confirmed this is what is happening

lsh-0:
raml changes are in and so are the code changes.

I'd like to merge this Friday if there are no further concerns

src/api.raml
```diff
@@ -715,6 +715,27 @@ traits:
                                     displayName: Complete
                                     value: !include samples/collection/v2/complete.json
 
+/bioprotocol:
```
giorgiosironi, line 718: mark with [`@experimental`](https://github.com/elifesciences/api-raml/blob/40b376afaa33083a9a86dd3516f171e091e7ffd0/adr/0002-use-raml-annotation-experimental.md) so that we can make amendments later without versioning

giorgiosironi (approved):

src/api.raml
```diff
@@ -715,6 +715,27 @@ traits:
...
+        get:
+            description: BioProtocol protocol data for a specific article
+            responses:
+                200:
+                    description: BioProtocol data
+                    body:
+                        application/vnd.elife.bioprotocol+json;version=1:
+                            schema: !include ../dist/model/bioprotocol.v1.json
+                            examples:
+                                complete:
```
thewilkybarkid, line 736: I'd add an `empty` example too.

src/model/bioprotocol.v1.yaml
```diff
@@ -0,0 +1,21 @@
+$schema: http://json-schema.org/draft-04/schema#
+title: BioProtocol
+type: array
+items:
+    type: object
```
thewilkybarkid: Everything should be `required`?

src/model/bioprotocol.v1.yaml
```diff
@@ -0,0 +1,21 @@
...
+    properties:
+        sequencing_number:
```
thewilkybarkid: Using camelCase elsewhere. Since this is a section ID, call it `sectionId`?

src/model/bioprotocol.v1.yaml
```diff
@@ -0,0 +1,21 @@
...
+            description: "section ID, looks like 's4-1'"
+            type: string
+        title:
```
thewilkybarkid: Not needed for machines, but probably useful for humans.

src/model/bioprotocol.v1.yaml
```diff
@@ -0,0 +1,21 @@
...
+            description: "section title, looks like 'Isolation of LDs and treatments'"
+            type: string
+        status:
+            description: "numeric code corresponding to the current status of a protocol"
+            type: "boolean"
```
thewilkybarkid: Numeric or boolean? If it a boolean, name needs to match (`status` doesn't work), otherwise an enum should be used.

lsh-0: it was changed to boolean from an integer. I'll change the description.

Why doesn't 'status' work? is it a reserved keyword or something?

thewilkybarkid: Doesn't work as in I don't know what `status: false` means (same with `status: 1`).

lsh-0: how about `has_protocol` ?

lsh-0: this boolean value indicates whether a 'request protocol' link should be created or a 'view protocol' link should be created. `status` seems fine to me, `has_protocol` doesn't seem so bad either

lsh-0: I'm keeping `status` and have improved the description. It's clearer now.

src/api.raml
lsh-0: done

src/model/bioprotocol.v1.yaml
lsh-0: everything is now required

src/model/bioprotocol.v1.yaml
lsh-0: done, is now camelCase

lsh-0: renamed sequencingNumber to sectionId

src/api.raml
lsh-0: done

lsh-0:
review feedback has been incorporated. the output structure more closely resembles the lists of results of everything else in the api now.

PR #242, opened 2020-03-23
Introduce promotional collections

nlisgo:
~Introduce regional collections~

External relations have made the decision that these should be promotional collections which can meet the requirements for their regional highlights. The are similar to collections. The main difference is that they have an optional list of associated editors.


src/api.raml
```diff
@@ -1711,6 +1711,66 @@ traits:
...
+        get:
+            description: |
+                Get a regional collection.
+            responses:
+                200:
+                    description: |
+                        Regional collection.
+                    body:
+                        application/vnd.elife.regional-collection+json;version=1:
+                            (experimental):
```
thewilkybarkid: This should be at the top level.

src/model/regional-collection.v1.yaml
```diff
@@ -0,0 +1,81 @@
...
+                                        enum:
+                                          - event
+                              - $ref: ../snippets/event.v1.yaml
+                          - allOf:
+                              - properties:
+                                    type:
+                                        type: string
+                                        enum:
+                                          - interview
+                              - $ref: ../snippets/interview.v1.yaml
```
thewilkybarkid: Any other types we might want to include?

src/model/regional-collection.v1.yaml
```diff
@@ -0,0 +1,81 @@
...
+            type: array
+            items:
+                $ref: '#/allOf/1/properties/content/items'
+            uniqueItems: true
+        podcastEpisodes:
+            title: Related podcast episodes
+            type: array
+            items:
+                $ref: ../snippets/podcast-episode.v1.yaml
+            uniqueItems: true
```
thewilkybarkid: Know these exist in normal collections, do we definitely need them here too?

src/model/regional-collection.v1.yaml
```diff
@@ -0,0 +1,81 @@
...
+  - properties:
+        image:
+            type: object
+            properties:
+                banner:
+                    $ref: ../misc/image.v1.yaml
+            required:
+              - banner
+        editors:
+            description: List of editors, ordered by name
```
thewilkybarkid: Is a different term needed here?

src/snippets/regional-collection.v1.yaml
```diff
@@ -0,0 +1,35 @@
...
+        properties:
+            thumbnail:
+                $ref: ../misc/image.v1.yaml
+        required:
+          - thumbnail
+    subjects:
+        description: Collection subjects
+        type: array
+        items:
+            $ref: ../snippets/subject.v1.yaml
```
thewilkybarkid: Is this needed?

nlisgo: I will ask, if not needed now I may still suggest that we keep it here but not implement it in Journal CMS at this moment. I suspect that we may need to support a listings page at some point for these collections.

src/model/regional-collection.v1.yaml
nlisgo: When we present these in the Journal it will be under the heading of Local editors.

src/model/regional-collection.v1.yaml
nlisgo: I will confirm but there was a desire to allow for support of all the options available with collections.

src/model/regional-collection.v1.yaml
nlisgo: I think if we need to add types here it will be inline with the types that are supported with collections. I will confirm the list is ok as it stands.

src/api.raml
nlisgo: Do you mean that we should not have an api (unused at this moment) for the regional-collection-list but just bump this up to the top level?

thewilkybarkid: Meant the experimental flag.

src/snippets/regional-collection.v1.yaml
nlisgo: Have confirmed that we should keep this in place to keep parity with collections but we may not expose this field at this moment in Journal CMS as the value will not be in use at this moment. It is an optional field in any case.

src/model/regional-collection.v1.yaml
nlisgo: After a discussion with Rowena, I can confirm that we should keep the same associations as are in place with collections.

src/model/regional-collection.v1.yaml
nlisgo: No other types are needed.

src/api.raml
```diff
@@ -1674,6 +1674,66 @@ traits:
...
+                        application/vnd.elife.promotional-collection+json;version=1:
+                            schema: !include ../dist/model/promotional-collection.v1.json
+                            examples:
+                                minimum:
+                                    displayName: Minimum
+                                    value: !include samples/promotional-collection/v1/minimum.json
+                                complete:
+                                    displayName: Complete
+                                    value: !include samples/promotional-collection/v1/complete.json
+
```
giorgiosironi, line 1736: the resource definition seems clean, now that the presence of subjects has been clarified

src/model/promotional-collection.v1.yaml
```diff
@@ -0,0 +1,79 @@
...
+                              - $ref: ../snippets/interview.v1.yaml
+            minItems: 1
+            uniqueItems: true
+        relatedContent:
+            title: Related content
+            type: array
+            items:
+                $ref: '#/allOf/1/properties/content/items'
+            uniqueItems: true
+        podcastEpisodes:
```
giorgiosironi, line 72: who knows why these are not considered `content`, but it fits https://github.com/elifesciences/api-raml/blob/develop/src/model/collection.v2.yaml so consistency is a priority.

giorgiosironi (commented):
I don't have the final word in this area, but I can't find any unaddressed issue. The use of @experimental safeguards additional changes.

nlisgo:
@thewilkybarkid is there anything left to do here or are you happy to approve?

thewilkybarkid (approved):

PR #243, opened 2020-04-15
Add tweet block

nlisgo:


nlisgo:
It is my understanding that, I will need to adjust the versions of all endpoints that support the new tweet block.

thewilkybarkid:
Technically this should have a version bump. However, it's probably ok to be pragmatic about it and add support everywhere before allowing content to be added containing it.

nlisgo:
> Technically this should have a version bump. However, it's probably ok to be pragmatic about it and add support everywhere before allowing content to be added containing it.

I could deploy the changes to Journal CMS last.

src/blocks/tweet.v1.yaml
```diff
@@ -0,0 +1,40 @@
...
+        title: Tweet ID
+        type: string
+    date:
+        $ref: ../misc/partial-date.v1.yaml
+    text:
+        title: Text
+        type: array
+        items:
+            $ref: paragraph.v1.yaml
+        minItems: 1
```
thewilkybarkid: Do we have paragraph blocks, or a single string of HTML?

nlisgo: > Do we have paragraph blocks, or a single string of HTML?

Single string of html.

PR #244, opened 2020-04-16
Add google map block

nlisgo:


src/blocks/google-map.v1.yaml
```diff
@@ -0,0 +1,29 @@
...
+            type: integer
+            minimum: 1
+        allowFullscreen:
+            type: boolean
+            default: true
+    required:
+      - type
+      - id
+      - width
+      - height
```
thewilkybarkid: Could have a default? (The ratio changing shouldn't make a difference with this type of embed.)

nlisgo: > Could have a default? (The ratio changing shouldn't make a difference with this type of embed.)

I believe that you're saying that we can expect that all google maps we wish to include will have the same aspect ratio so there is no need to calculate it? This would make the Journal CMS simpler also and I can drop the parameters here.

thewilkybarkid: You could add in optional width/height later, but unlike embedding a YouTube video, for example, Google Maps should be able to work at different aspect ratios.

PR #246, opened 2020-05-14
Add structured abstract and bump appropriate versions

nlisgo:
Although abstracts are not present in most of the sampled content it is an optional field and may be present so this one change is impacting quite a lot of associated content.


lsh-0:
oof - this is a lot of changes, I didn't realise it would have such far reaching extent.

I probably won't have time to review this today (Fri 29th). 

I'm also worried the number of changes will overwhelm specific sections we should be paying attention to. Were they are any particular changes you think we should give extra scrutiny to @nlisgo ?

@thewilkybarkid , I appreciate you're in a separate team now but your feedback would be especially welcome too.

lsh-0:
@nlisgo , I've added @gnott to the list of reviewers and he's also working on structured abstracts right now and may be able to give you some feedback faster than I can.

src/misc/article.v3.yaml
```diff
@@ -0,0 +1,85 @@
+$schema: http://json-schema.org/draft-04/schema#
```
gnott, line 1: This file looks pretty much the same as the `src/misc/article.v2.yaml` file, is it necessary to add this new one, or maybe I don't understand fully?

src/snippets/article.v2.yaml
```diff
@@ -0,0 +1,118 @@
...
+                          content:
+                              description: Content
+                              type: array
+                              items:
+                                  $ref: ../blocks/paragraph.v1.yaml
+                              minItems: 1
+                      - $ref: ../blocks/paragraph.v1.yaml
+                minItems: 1
+        required:
+          - content
```
gnott: To not reuse the standard `section` block here looks like a good idea, to restrict the allowed content, at least I hope it is. The XML sample I have has some `@id` attributes on the `<sec>` tags, I wonder if adding an optional `id` attribute for the section might be something useful?

src/misc/article.v3.yaml
nlisgo: There's one difference and that is the reference to `- $ref: ../snippets/article.v2.yaml` rather than `- $ref: ../snippets/article.v1.yaml`. I believe it may be possible to inherit from `src/misc/article.v2.yaml` and just articulate the changes from that file but I can't find another example of that approach in this project. Although that approach may make the code reviews easier it may make it more difficult to know the current status of the schema as each version just contains a delta.

src/snippets/article.v2.yaml
nlisgo: I will add this.

src/misc/article.v3.yaml
gnott: Thanks, it sounds like technically possible but it's different for stylistic reasons, sounds fine to me, and it clarifies things for me.

gnott: Oh, I think I see my error, and I see the necessity to bump it now

gnott:
I may be wrong, but would `src/model/digest-list.v1.yaml` also get bumped to a `v2`, and it references `$ref: ../snippets/digest.v2.yaml`?

src/samples/article-vor/v4/unpublished-complete.json
```diff
@@ -0,0 +1,1642 @@
...
+    },
+    "abstract": {
+        "doi": "10.7554/eLife.09560.001",
+        "content": [
+            {
+                "type": "paragraph",
+                "text": "This is an example of an article with a structured abstract."
+            },
+            {
+                "type": "section",
```
gnott, line 330: May be able to add an `"id"` here when the defintion allows it?

gnott (approved):
With my level of understanding, these changes look pretty good, and represent some very detailed and  labour intensive updates - looks really good @nlisgo!

I don't know how much work is required on other projects to use these or when it is safe to merge, but I can leave that to those who know more about those parts. If there's something additional I could help with to integrate, I'm open to helping out.

nlisgo:
> I may be wrong, but would `src/model/digest-list.v1.yaml` also get bumped to a `v2`, and it references `$ref: ../snippets/digest.v2.yaml`?

Although the digest list that we prepare contains the whole digest, as far as I can tell, the schema actually declares only a partial digest. The `relatedContent` property is not declared in the schema for the digest list but we put it in there. If digest is the only consumer of the digests we should perhaps adjust the output of digest list.

nlisgo:
@thewilkybarkid although you are not a maintainer on this your input would be valuable and appreciated. This is the biggest single change I have introduced to `api-raml`.

I would like to understand why abstract is included in listings. Do we have examples on the journal site of the abstract being utilised within a listing? If not, I believe we could simplify this quite significantly.

nlisgo:
I believe that all feedback from @gnott has now been addressed.

thewilkybarkid:
> @thewilkybarkid although you are not a maintainer on this your input would be valuable and appreciated. This is the biggest single change I have introduced to `api-raml`.
> 
> I would like to understand why abstract is included in listings. Do we have examples on the journal site of the abstract being utilised within a listing? If not, I believe we could simplify this quite significantly.

Abstracts need to be in the recommendations API for the the 'further reading' list on article (see also https://github.com/elifesciences/issues/issues/4704).

Options I can see:

1. Remove abstracts from snippets entirely (Journal will need to fetch the articles to get them; handful of extra requests, but with caching shouldn't be noticable)
2. Remove abstracts from snippets unless they're needed (Recommendations could fetch the articles to get them, or they also remain in the APIs that Recommendations requests from; makes the APIs a bit inconsistent)
3. Don't do a version bump (not following semver, but safe if you do it in the right order as there are no known external clients)
4. Do the version bump

nlisgo:
@thewilkybarkid thanks for the feedback.

nlisgo:
It's tempting to take the following approach. Remove the abstract from the article snippets without a version bump. This will mean that we have to perform an additional request for each of the recommendations that we display on the journal. I think this will not have a big impact. 

Once this is in place I will adjust this PR and the diff should shrink significantly.

Thoughts @lsh-0 @gnott?

gnott:
> Remove the abstract from the article snippets without a version bump

Now knowing recommendations is the primary consumer of abstracts in snippets, and it sounds like not too many other services are likely to use the snippet abstract, it would be a good simplification to the schema to remove abstract from snippets.

Of course, these specific changes do not affect any of the projects I maintain, and I only know the other journal parts at a high level, so I may be missing important details about other impacts.

lsh-0:
> but with caching shouldn't be noticable

where is this caching coming from? `api-sdk-php`? if these are internal calls between microservices that don't leave the VPC then there is no CDN involved.

I'm leery of data removals without a formal change to the spec version. With a proper change to the version of a spec Lax can downgrade or upgrade content based on the version requested. For example, if a v4 article snippet is requested (that has no abstract) then I can take the data I already have, drop the abstract and serve up valid v4 content.

We also have a proper excuse then when applications start failing left and right because they rely on  content that is now missing: *"you should be specifying the version of the content you want"*.

Applications that fail noisily are also easier to find and fix.

I'm also a fan of minimal network overhead, so fewer HTTP requests.

Bending the rules now to introduce exceptions to the rule and sneaky changes that make our lives more convenient is pure technical debt.

I recommend we do this properly, bump the versions and soldier on.

It sounds outrageous right now, but if the api-raml becomes too onerous to deal with at the pace we want to work at, we can investigate how to replace it.

lsh-0:
Decision is ultimately Nathan's. Let us know what you decide @nlisgo

nlisgo:
As abstract is currently an optional field it would not be breaking to remove it from the current versions. I will work on a PR to remove the abstract from the schema of all listings except recommendations. This would mean that in the one instance we need it in a listing, we avoid additional queries. If we agree to adopt those changes then I will adapt this PR and version bump all endpoints impacted by the structured abstract change.

nlisgo:
PR submitted which strips the abstract from all listings except `recommendations`: https://github.com/elifesciences/api-raml/pull/248

We don't need to prioritise omitting the value from listings that are already serving up the abstract, as serving up properties that are not declared in the schema does not violate schema checks. 

Once that PR is merged in, this PR will be updated and will contain version bumps for impacted endpoints but the changes should be significantly smaller.

nlisgo:
Once https://github.com/elifesciences/api-raml/pull/248 has been merged the structured abstract introduction will just affect 3 endpoints. See this comparison: https://github.com/elifesciences/api-raml/compare/remove-abstract-from-snippet...structure-abstract-2

lsh-0:
Added a summary in that PR: https://github.com/elifesciences/api-raml/pull/248#issuecomment-642419444-permalink

Implementation wise it shouldn't be too difficult. 

If the abstract in the snippet was optional then it's not unreasonable for it to go away and expect code to continue work. Code isn't always written to the spec though so I'm still expecting downstream effects when lax stops including it.

Regarding backwards compatibility, when a request is made for a VOR spec-version 4 article that is using a structured abstract but their request only supports content-types of VOR spec-version 3, is there a way to translate the structured abstract into a regular abstract and successfully fulfill the request? Or should I just fail the request with a [HTTP 415](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#415) ? 

I've been able to downgrade content previously for spec-version 1 and 2 article content.

Could I elide the abstract altogether so that it validates against a VOR spec-version 3? (assuming the abstract within an article is optional, which I'm unsure about).

lsh-0 (approved):
approved, assuming I haven't completely misunderstood everything.

nlisgo:
> Regarding backwards compatibility, when a request is made for a VOR spec-version 4 article that is using a structured abstract but their request only supports content-types of VOR spec-version 3, is there a way to translate the structured abstract into a regular abstract and successfully fulfill the request? Or should I just fail the request with a HTTP 415 ?

I think we should just fail the requests. It's less code to maintain simply to detect that there is a structured abstract and then respond with the appropriate status code. We may need to compare notes on that because I've been using [HTTP 406](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#406).

Maybe we should decide to only maintain 2 versions maximum for each media type and mark the one below the latest as deprecated and remove any earlier ones. We can discuss this at some point.

nlisgo:
Having merged in https://github.com/elifesciences/api-raml/pull/248 this PR is a lot cleaner. I'm inviting review again.

lsh-0:
> We may need to compare notes on that because I've been using HTTP 406.

I just grepped through the lax code and I can see usage of both. I'd need to take a closer look but that can wait until I implement support for this

> Maybe we should decide to only maintain 2 versions maximum for each media type and mark the one below the latest as deprecated and remove any earlier ones.

yes, I think lax still supports VOR article specs 1, 2, and 3. Requests for spec version 1 get a deprecated warning in the header or something. It's blurry, it's been a while.

src/api.raml
```diff
@@ -442,6 +458,27 @@ traits:
...
+                                00569:
+                                    displayName: '00569'
+                                    value: !include samples/article-vor/v3/00569.json
+                                15278:
+                                    displayName: '15278'
+                                    value: !include samples/article-vor/v3/15278.json
+                        application/vnd.elife.article-vor+json;version=4:
                             (experimental):
                             schema: !include model/article-vor.v3.json
                             examples:
```
lsh-0, line 484: do these examples need to be updated from

    !include samples/article-vor/v3/minimum.json

to

    !include samples/article-vor/v4/minimum.json

? It looks like you have the files included in the PR

src/samples/article-vor/v4/00569.json
```diff
@@ -0,0 +1,7679 @@
...
+        {
+            "id": "developmental-biology-stem-cells",
+            "name": "Developmental Biology and Stem Cells"
+        },
+        {
+            "id": "genomics-evolutionary-biology",
+            "name": "Genomics and Evolutionary Biology"
+        }
+    ],
+    "abstract": {
```
lsh-0, line 26: should we have an article with a structured abstract as an example of a valid v4 article?

lsh-0 (approved):
I think the examples may need their paths changed but apart from that it looks fine

nlisgo:
> yes, I think lax still supports VOR article specs 1, 2, and 3. Requests for spec version 1 get a deprecated warning in the header or something. It's blurry, it's been a while.

I will create a ticket to introduce the proposed change to api-raml and the PR can be the discussion thread. We could update the README at the same time to make it clear what the plan is for maintaining older versions. Journal CMS is serving previous versions.

src/samples/article-vor/v4/00569.json
nlisgo: > should we have an article with a structured abstract as an example of a valid v4 article?

I've added one now.

nlisgo:
> I think the examples may need their paths changed but apart from that it looks fine

I've fixed this.

lsh-0:
cheers, @nlisgo

PR #248, opened 2020-06-09
Remove abstract from article snippet except in recommendations

nlisgo:
The important removals are found here:
- https://github.com/elifesciences/api-raml/pull/248/files#diff-9e840b084d9f1ea1e0e92abfc57ea327
- https://github.com/elifesciences/api-raml/pull/248/files#diff-badbd555d4ea1d981b262d2b50ab3b4b

Removing the abstract property from the snippets strips the property from all listings. Because the full article models borrow from the snippet then we have to declare the property here:
- https://github.com/elifesciences/api-raml/pull/248/files#diff-88a590b4913ac35e067c01ee1d20d8b0
- https://github.com/elifesciences/api-raml/pull/248/files#diff-0351a4a0208c0ca05d9c16de3cc955f9

As recommendations is the only listing that we need the abstract we can declare the abstract here:
- https://github.com/elifesciences/api-raml/pull/248/files#diff-9340409dc67ddc2522b0fd9c72a99d61

### DIFF
--- a/src/model/article-vor.v1.yaml
+++ b/src/model/article-vor.v1.yaml
@@ -5,6 +5,11 @@ allOf:
   - $ref: ../snippets/article-vor.v1.yaml
   - $ref: ../misc/article.v1.yaml
   - properties:
+        abstract:
+            type: object
+            properties:
+                doi:
+                    $ref: ../misc/doi.v1.yaml
         keywords:
             type: array
             items:
--- a/src/misc/abstract.v1.yaml
+++ b/src/misc/abstract.v1.yaml
@@ -1,0 +1,11 @@
+$schema: http://json-schema.org/draft-04/schema#
+title: Abstract
+type: object
+properties:
+    content:
+        type: array
+        items:
+            $ref: ../blocks/paragraph.v1.yaml
+        minItems: 1
+required:
+  - content
--- a/src/misc/article.v1.yaml
+++ b/src/misc/article.v1.yaml
@@ -4,6 +4,8 @@ type: object
 allOf:
   - $ref: ../snippets/article.v1.yaml
   - properties:
+        abstract:
+            $ref: abstract.v1.yaml
         issue:
             type: integer
             minimum: 1
--- a/src/misc/article.v2.yaml
+++ b/src/misc/article.v2.yaml
@@ -4,6 +4,8 @@ type: object
 allOf:
   - $ref: ../snippets/article.v1.yaml
   - properties:
+        abstract:
+            $ref: abstract.v1.yaml
         xml:
             type: string
             format: uri
--- a/src/model/article-vor.v2.yaml
+++ b/src/model/article-vor.v2.yaml
@@ -5,6 +5,11 @@ allOf:
   - $ref: ../snippets/article-vor.v1.yaml
   - $ref: ../misc/article.v2.yaml
   - properties:
+        abstract:
+            type: object
+            properties:
+                doi:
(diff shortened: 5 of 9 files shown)


gnott:
I think we recently made it so abstract DOI is optional, in case that applies here.

nlisgo:
> I think we recently made it so abstract DOI is optional, in case that applies here.

Abstract DOI is optional for VOR articles. But, if the property is present then it will perform the following regex on the value: `^10[.][0-9]{4,}[^\s"/<>]*/[^\s"]+$`

nlisgo:
I've updated the description.

nlisgo:
This comparison view give us an idea of what the structured abstract PR will look like once we merge this in: https://github.com/elifesciences/api-raml/compare/remove-abstract-from-snippet...structure-abstract-2

src/model/article-vor.v1.yaml
```diff
@@ -5,6 +5,11 @@ allOf:
   - $ref: ../snippets/article-vor.v1.yaml
   - $ref: ../misc/article.v1.yaml
   - properties:
+        abstract:
+            type: object
+            properties:
```
gnott, line 10: I may be wrong, but it looks like the `doi` property itself remains optional in an abstract, and can the `doi` property just move to the `src/misc/abstract.v1.yaml` file? That may imply a PoA abstract can also have an optional `doi` property, which doesn't seem too much of a stretch of the rules. Then, wouldn't this file get its abstract from the `src/misc/article.v2.yaml ` file referenced?

gnott:
To me this update seems possible, as you say @nlisgo, because the `abstract` is an optional property on all articles, then `abstract` is added where it is useful for recommendations. I'm not the person to ask whether this is all good as written for adapting existing model definitions, but to my eyes it's a very smart shifting of values.

lsh-0:
alright - my understanding of the current state is this:

1. we remove the article abstract from article snippets (the summary article information) which means they also disappear in snippet listings, *except* for the 'recommendations' listing. Otherwise, the journal would have to do further requests for article data to find the abstract.

2. we introduce a new 'abstract' spec with versions 1 and 2. Abstract spec version 1 is what we currently use and spec version 2 are the new structured abstracts.

3. we introduce a new POA version 3 spec and a VOR version 4 spec that adds support for structured abstracts.

4. we introduce a new 'recommendations' version 2 spec that adds support for structured abstracts.

This set of changes affects the `lax` and `recommendations` services.

Is my understanding correct?

I suspect there are applications that are pulling from the API and not specifying content versions in their requests still under the assumption the abstract will always be there and in a certain form. I hope they fail noisily during testing.

lsh-0 (approved):

src/model/article-vor.v1.yaml
nlisgo: I would like to retain the same rules for the abstract when it is present so I'm not going to introduce this now. It may be worth considering but I don't want to question the decision to not make doi optional for POA as part of this PR. Although the diff is a bit bigger than it need be if I made this adjustment it would make it, it's not substantial.

Thanks for your feedback.

nlisgo:
> Is my understanding correct?

Yes!

> I suspect there are applications that are pulling from the API and not specifying content versions in their requests still under the assumption the abstract will always be there and in a certain form. I hope they fail noisily during testing.

Journal is the main consumer of the API and it uses the `api-sdk-php` project to model the data returned from the api requests. I can verify that abstracts are treated as optional for `poa` and `vor` articles and recommendations.

PR #251, opened 2020-06-19
/article deprecations and obsolesence

lsh-0:
* removes poa v1 and vor v1. 
* deprecates vor v2

PR #252, opened 2020-06-26
removes article vor v2

lsh-0:
Follow up work from: https://github.com/elifesciences/api-raml/pull/251

- removes article vor v2
- deprecates article poa v2
- deprecates article vor v3
- drops 'experimental' against article poa v3
- drops 'experimental' against article vor v4


lsh-0:
no urgency on this one, @nlisgo , it's just some clean up work to ensure we only support two versions of an article spec at a time.

nlisgo (approved):

lsh-0:
thanks for the review, @nlisgo . I'll take a look at what is required in lax tomorrow

lsh-0:
I've taken a look at lax (https://github.com/elifesciences/lax/pull/537) and how it's referencing the schema and because we're not actually removing the earlier article models like poa or vor v1 etc, just support in the /article endpoints for those obsolete types, then this should be fine to merge.

PR #253, opened 2020-10-02
Doi social media image

nlisgo:
@lsh-0 We would need to expose another article store endpoint similar to `/articles/[articleId]/fragments/image` which would allow us to add, modify or delete an article `socialImage`.

This would need to be available regardless of whether an article was `vor` or `poa`. 

As this new property is optional we shouldn't need to bump versions.

### DIFF
--- a/src/misc/article.v3.yaml
+++ b/src/misc/article.v3.yaml
@@ -4,6 +4,8 @@ type: object
 allOf:
   - $ref: ../snippets/article.v1.yaml
   - properties:
+        socialImage:
+            $ref: image.v1.yaml
         abstract:
             $ref: abstract.v2.yaml
         xml:
--- a/src/samples/article-poa/v3/complete.json
+++ b/src/samples/article-poa/v3/complete.json
@@ -149,8 +149,8 @@
             "filename": "an-image.jpg"
         },
         "size": {
-            "width": 4194,
-            "height": 4714
+            "width": 1772,
+            "height": 1782
         },
         "focalPoint": {
             "x": 25,
--- a/src/samples/article-poa/v3/complete.json
+++ b/src/samples/article-poa/v3/complete.json
@@ -137,6 +137,26 @@
         "Human",
         "Xenopus"
     ],
+    "socialImage": {
+        "uri": "https://iiif.elifesciences.org/lax/14107%2Felife-14107-fig1-v3.tif",
+        "alt": "",
+        "attribution": [
+            "By some person."
+        ],
+        "source": {
+            "mediaType": "image/jpeg",
+            "uri": "https://iiif.elifesciences.org/lax/14107%2Felife-14107-fig1-v3.tif/full/full/0/default.jpg",
+            "filename": "an-image.jpg"
+        },
+        "size": {
+            "width": 4194,
+            "height": 4714
+        },
+        "focalPoint": {
+            "x": 25,
+            "y": 75
+        }
+    },
     "abstract": {
         "content": [
             {
--- a/src/samples/article-vor/v4/complete.json
+++ b/src/samples/article-vor/v4/complete.json
@@ -346,6 +346,26 @@
             }
         }
     },
+    "socialImage": {
+        "uri": "https://iiif.elifesciences.org/lax/09560%2Felife-09560-fig1-v1.tif",
+        "alt": "",
+        "attribution": [
+            "By some person."
+        ],
+        "source": {
+            "mediaType": "image/jpeg",
+            "uri": "https://iiif.elifesciences.org/lax/09560%2Felife-09560-fig1-v1.tif/full/full/0/default.jpg",
+            "filename": "an-image.jpg"
+        },
+        "size": {
+            "width": 4194,
+            "height": 4714
+        },
+        "focalPoint": {
+            "x": 25,
+            "y": 75
+        }
+    },
     "abstract": {
         "doi": "10.7554/eLife.09560.001",
         "content": [


lsh-0:
where is the issue for this?

> This would need to be available regardless of whether an article was vor or poa.

This sounds like you've come up with a solution before fully exploring the problem or it's consequences. 

What is needed (it sounds like, again, no link to issue or requirements) is that articles need additional data attached to them, like images. Lax is not the right place for this. The journal-cms is supposed to be the place where regular folk come to manage content for the journal that isn't tied to our strict production processes. If anything I would think the journal cms is *perfectly* suited to this. 

If I've misunderstood the purpose of the journal-cms please correct me.

I'd be grateful for a link to the issue where this became a requirement.

Also, use of `fragment` here in a RESTful URL is bad form:

* articles don't have fragments
* Lax may have unused functionality that uses the term 'fragment', but that is an implementation detail
* 'fragment' within a URL is already a [loaded term](https://en.wikipedia.org/wiki/URI_fragment)

cc @NuclearRedeye

lsh-0:
is this the issue? https://github.com/elifesciences/issues/issues/6160

I propose a formal process for modifying the API. It would cover any modifications to the API, like adding new endpoints and removing or deprecating content types. It would also explore the affected projects and the changes that would have to be made.

I'd like whatever the issue was that created this work to be the first to follow this process.

NuclearRedeye:
@lsh-0 [This](https://github.com/elifesciences/issues/issues/5710) is the overall project that spawned the above issue and PR.

> Lax is not the right place for this.

I can't comment if it is the right place or not, but @nlisgo thinking if I recall was that I think the 'interesting image' already works this way, and as such his thoughts were that this should just work the same way. If that isn't the case however, maybe it means that the 'interesting image' also needs to possibly be re-located?

>  propose a formal process for modifying the API.

Sounds sensible, was there a process that was followed before?

lsh-0:
> Sounds sensible, was there a process that was followed before?

we begged Chris, which was both a blessing and a curse. It's a curse having a gatekeeper because the speed of changes are throttled, but also a blessing because the changes are well considered and then coordinated (Chris was across most/all PHP projects).

Changes right now are adhoc and a bit chaotic. We can trace the current problems with `search` back to it. A formal process would replace Chris as the gatekeeper but keep changes coordinated. We would have an opportunity to review proposals and tweak decisions before work went ahead - especially by project owners whose projects will be affected.

nlisgo:
Journal CMS is absolutely the place to upload the image. I envisaged that it would behave much like the interesting image for article VOR. Lax exposes an endpoint and Journal CMS acts as a client to allow our team to upload the image but lax being the source of truth for articles stores and returns the image in the article json response.

nlisgo:
> I'd like whatever the issue was that created this work to be the first to follow this process.

I don't mind if this issue is the catalyst for introducing a formal process. I think we should jump on a call this week sometime to agree on a way of addressing this within this sprint. It seems that you are unhappy with the implementation of the interesting image, since Journal CMS is the only service to interface with this, I don't mind if we agree a better approach for both as part of the same piece of work. I would not add too much on my end. We can remove the reference to `fragment` in the endpoint url. We all understand that the intent was to allow Journal CMS to introduce the `image` only and not risk accepting any wider changes to the article data.

> Changes right now are adhoc and a bit chaotic. We can trace the current problems with search back to it.

This is loaded. Structured abstracts have been more of a pain than they were anticipated to be for sure.

lsh-0:
>  It seems that you are unhappy with the implementation of the interesting image,

Incorrect, I'm unhappy about implementation of this going ahead without being consulted. I don't like having to shoot down your work or become a bottleneck and I don't like being left out of the design process. I don't like having to scramble to keep up with and make sense of decisions that will affect the running of otherwise very stable and important systems.

> > Changes right now are adhoc and a bit chaotic. We can trace the current problems with search back to it.

> This is loaded. Structured abstracts have been more of a pain than they were anticipated to be for sure.

No, it's not loaded. It isn't a once-off, it's a pattern of behaviour that arises from siloed developers making decisions unilaterally. It happened with structured abstracts, happened with ERA/RDS and it is happening again now. A large amount of infrastructure is composed of tightly coupled PHP applications and libraries that are sensitive to API changes most of which are now abandoned and unmaintained. We change one thing and are surprised when a bunch of other things stop working. It's why I have an open-ended ticket in this sprint to become a php developer and figure out why `search` is so hosed.

I'll take a look at the ticket @NuclearRedeye linked to today and leave some feedback.

nlisgo:
> Incorrect, I'm unhappy about implementation of this going ahead without being consulted.

You have referenced use of the word `fragment` in the endpoint url as being undesirable. This is not a requirement, I only referenced the endpoint of the existing API for the interesting image to illustrate that a similar approach may be needed for this social media image.

The PR for a schema change is a form of communication and at the moment since we are in different time zones may be the best and sometimes first way to articulate some of the changes that we need to serve new requirements. If you want earlier intervention then they tend to be in meetings and google documents and then they finally arrive in the sprint as priorities. I'm available most days to hang on after a stand up call but you haven't once made such a request.

I was being facetious when I said "This is loaded". I apologise for that. I was making light of your frustrations but I should leave that out of this style of communication.

In all of this, you have not given me a shred of hope that this PR is close to your consideration. I have no feedback to address relevant to the schema of the article to accommodate a social media image.

lsh-0:
I've taken a look at the original ticket and the changes you're introducing here.

> @lsh-0 We would need to expose another article store endpoint similar to /articles/[articleId]/fragments/image which would allow us to add, modify or delete an article socialImage.
> This would need to be available regardless of whether an article was vor or poa.

If I understand the intent of what you are doing, you are relying on your memory of the 'fragment' system in lax to introduce snippets of article data that are not present during production. Some important points:

1. nothing is using the fragment system, it is untested in the real world.
2. this will create a new coupling between journal-cms and lax. Less, not more, coupling is better.
3. attaching random images and other media to an article, either one or many or none, seems like the job of a content management system. I don't know why journal-cms isn't doing this.

Some details of the fragment system:

* It attaches snippets of json to an article across *all* versions of an article, and not a specific version.
* Failure to validate will see the snippet rejected
* The snippet must have an identifier so you can HTTP DELETE it later
* I don't recall the merging mechanism, you may be limited to where in the article you can add your data 

I still want to formalise this process given I've just spent the entire day beating my head against `search` and hacky code, and getting nowhere. For now I'm dragging @gnott in for his thoughts as well.

nlisgo:
The decision to store the interesting image with lax set a precedent which is obviously affecting my approach here. I'm quite willing to revisit whether the interesting image should have just lived with Journal CMS. With interesting image this would present more of a challenge than the social media image as we introduce it in the listing.

But, I'm not going to take on the interesting image now. I will mark this PR as WIP just to invite thoughts from @gnott and any other feedback.

I have just had a discussion with @NuclearRedeye and have suggested a different approach for the social image and I will move in that direction a little and raise a WIP PR so we can have a discussion about the merits/pitfalls of that approach.

nlisgo:
See: https://github.com/elifesciences/api-raml/pull/254

for an alternative approach for the article and magazine content social media image. Please add your comments, feedback on that PR.

gnott:
The proposed changes to the schema itself, to me, look fairly logical. However, I think how things are current run and connected together it poses difficulty.

For example, during an article ingestion, `lax` + `bot-lax-adaptor` will accept the article XML, and then continues to fill out details about images (from IIIF), and videos (from Glencoe), and other modifications. Once the ingestion is completed, my understanding is only the publication dates and article status is modified when the article version is published. The article content will not change until that article is ingested again, either set in motion by an article ingestion workflow, or Lax re-generates data based on the corpus of article XML from `elife-article-xml` repo.

I wonder what would happen if someone edits the image or the image focal point in Journal CMS; would the expectation be that it would appear on journal? I think, based on the processes we have right now, it would need to be re-ingested by Lax each time the image is edited. Do we already have a way to ask Lax to update all an article's versions again on a per article basis? If we do (or if we planned to do that), would it be a good pattern to follow?

One alternative might be to borrow something from the related article / external article linkages. Those possibly only cover one use case which is linking article snippets together, and that might be too constrictive, if images will also relate to non-research-article content.

Digest endpoint currently runs as a separate app endpoint, and there it is a unidirectional one-to-many link from digest to article version; one digest relates to all versions of an article, by DOI / snippet data. I guess that is a little more like the alternative you have in https://github.com/elifesciences/api-raml/pull/254 @nlisgo, where the images are a top level structure. That might end up being a simple way to store the data and only resulting in one addiitonal API call to collect the content and image together for display on journal.

I hope I understood the basics of this feature when describing the above, based on what I generally understand about how things fit together.

lsh-0:
> I hope I understood the basics of this feature when describing the above, based on what I generally understand about how things fit together.

Your description above is correct. 

I've lost sleep over this the last few days wondering if I'm being too controlling and stubborn, but just in reading @gnott's comment I realised there are downstream effects I wasn't aware of.

I think it *could* be done in lax even with the caveats and unknowns mentioned, but I'm also thinking about the other areas we might need to 'stretch' to in the future (no reason to stop at an 'interesting image') and the best home for all these additional snippets of article data. This is where a better design process for API changes could be applied - are there are other changes in the pipeline besides the 'interesting image' that would improve whatever decision we make now to accommodate this change? 

I think @nlisgo's [new PR](https://github.com/elifesciences/api-raml/pull/254) exemplifies this: why stop at social media images for articles? Digests and labs and podcasts would benefit as well. But then would each of the individual services be updated to accommodate this (lax, digests, ... profiles?) or just the one service (journal-cms)?

I don't have cut and dried answers to all these questions I'm proposing, API design isn't *easy*, but performance also needs to be considered. If we're generating 2.2 million requests a month internally, can we predict what the additional load will be for one more API call by the journal?

These are decisions I don't want to make *myself* unilaterally or freak out about when forced to decide on the fly by approving a PR out of the blue.

lsh-0:
Looking at the two proposals (*this* one and [this one](https://github.com/elifesciences/api-raml/pull/254)), I'm erring on the side of this one. Introducing an `/interesting_images` endpoint seems far too specific but also more load on `journal-cms`, the API and more code in general.

@nlisgo , in the [parent ticket](https://github.com/elifesciences/issues/issues/5710) you've already specced out DODs that involve me updating Lax to accommodate the changes, but was it considered making the social image part of the production process?

`digests` is not much more than a subset of article data from lax already but with the addition of an image, which is essentially the same sort of image we're talking about here. If we were to add social image data to lax, digests would become redundant, which I think is funny because `digests` strikes me as an attempt to avoid modifying `lax` ... (I know that's not the whole story, but still).

Assuming we go with this option, are there plans to add this snippet to other content types? or do they already have equivalents?

nlisgo:
> @nlisgo , in the [parent ticket](https://github.com/elifesciences/issues/issues/5710) you've already specced out DODs that involve me updating Lax to accommodate the changes, but was it considered making the social image part of the production process?

I don't believe this was considered, I would need to raise this question and report back.

> digests would become redundant

If we did consider making the social image part of the production process it would not make digest redundant as I have already floated the idea of using the digest image for the doi content and this was not deemed an appropriate image in most cases. There may be some cases where the same image would be sufficient but we would just duplicate the images in this scenario.

> Assuming we go with this option, are there plans to add this snippet to other content types? or do they already have equivalents?

Yes, we will need a social media image for some (maybe most, checking this requirement now) of the magazine (Journal CMS) content.

lsh-0:
> If we did consider making the social image part of the production process it would not make digest redundant as I have already floated the idea of using the digest image for the doi content and this was not deemed an appropriate image in most cases. 

that is interesting, and would be easy to do for the RSS feeds once I add digests support...

nlisgo:
> I wonder what would happen if someone edits the image or the image focal point in Journal CMS; would the expectation be that it would appear on journal? I think, based on the processes we have right now, it would need to be re-ingested by Lax each time the image is edited.

As in the case of the interesting image we would need the ability to remove or modify the image. Having not had a role in the implementation of the lax interesting image `/articles/[articleId]/fragments/image` endpoint I was not aware of the specific challenges of this approach.

lsh-0:
> not aware of the specific challenges of this approach.

More like *constraints*. The fragments system uncovered some complex edge cases that I needed to strictly control for. Because it's never been used my confidence in it isn't great, I'll need some time to revisit it. I do suspect there is an end2end test for it though which would be helpful. I'll take a look tomorrow.

nlisgo:
@lsh-0 if we decide that Journal CMS is the preferred place to upload the image for all content (including DOI) then I think the best solution would something along the lines of https://github.com/elifesciences/api-raml/pull/254. This way we can avoid being reliant on the fragment api. We can do some work to measure the impact of the additional API call.

lsh-0:
Lets go with Lax and fragments. I'm not keen on the solution but we're going to have look at performance issues soon with our other services (like `recommendations`) and reducing the number of API calls is an easy win.

I'll need some time to review what is in Lax and get back to you with the API endpoints and their behaviour. I'll create a ticket for it and hopefully get it done tomorrow

lsh-0:
ticket here: https://github.com/elifesciences/issues/issues/6199

if @nlisgo and @NuclearRedeye could figure out it's priority I'd be grateful.

nlisgo:
Marking this as ready for review. Will close https://github.com/elifesciences/api-raml/pull/254.

lsh-0 (approved):
looks like a stock standard `image` section

PR #254, opened 2020-10-06
Interesting images endpoint v1

nlisgo:
This is an alternative proposal to allow the association of a social media image with doi and magazine content. 

Initially this could just display the socialImage if available but it could be expanding to handle any other images such as the DOI interesting image.


lsh-0:
Thanks for your work on this @nlisgo , it's good to see the options laid out like this. 

And the alternative for reference: https://github.com/elifesciences/api-raml/pull/253

PR #255, opened 2020-10-14
Refactor the social media image to introduce it to the images property

nlisgo:
@lsh-0 after merging in https://github.com/elifesciences/api-raml/pull/253 and beginning work on the social media image for other content types I think it may be a better option to just add the `social` image into the `image` property. We have examples where we store multiple images in the `image` property for other content: https://github.com/elifesciences/api-raml/blob/develop/src/samples/collection/v2/complete.json#L7-L42

This PR handles this refactor and adjusts the logic to ensure that if the `image` property (which is optional) is present in `article-vor` items within listings then the `thumbnail` property must be present. It the `image` property is present for a specific article version then either the `thumbnail` or `social` properties must be present. I have added samples for these cases.

I think this is the right approach for the magazine content but if you would prefer to have it as it's own top-level property as was introduced https://github.com/elifesciences/api-raml/pull/253 then we can just take a different approach for the doi content.


lsh-0 (approved):
Good catch and a better fit.

Give @gnott a chance to take a look and leave his review as well.

gnott:
The naming looks good, and I expect it will have no impact on generating JSON from XML since that attribute is a fragment? Approve!

gnott (approved):

lsh-0:
>  I expect it will have no impact on generating JSON from XML since that attribute is a fragment?

correct, it bypasses bot-lax-adaptor.

nlisgo:
Thanks for the quick review. I will incorporate these changes into the `api-dummy` PR: https://github.com/elifesciences/api-dummy/pull/170

src/snippets/article-vor.v1.yaml
```diff
@@ -19,5 +19,8 @@ allOf:
             properties:
                 thumbnail:
                     $ref: ../misc/image.v1.yaml
-            required:
-              - thumbnail
+            anyOf:
```
nlisgo, line 22: This should not have changed. This was an error. Need to work on this some more.

PR #256, opened 2020-10-15
Add social image to magazine content

nlisgo:
This PR shares the same challenge as `article-vor`. `image` is optional but when `image` is present is listing we expect the `thumbnail` property to present and if `image` is present on a full interview it should contain either the `social` or `thumbnail` property.

See: https://github.com/elifesciences/api-raml/pull/257


src/model/labs-post.v2.yaml
```diff
@@ -4,6 +4,11 @@ type: object
 allOf:
   - $ref: ../snippets/labs-post.v1.yaml
   - properties:
+        image:
+            type: object
+            properties:
+                social:
+                    $ref: ../misc/image.v1.yaml
```
NuclearRedeye, line 11: Just to double check, the social image is optional here right?

NuclearRedeye (approved):
Yup, look good to me. Just that one comment to double check that you intended to not have a `required` line.

src/model/labs-post.v2.yaml
nlisgo: Because the image is required in the `labs-post` snippet and 'thumbnail` is required, `social` is not stated as required. `social` should always be considered optional. It only is included in required if `image` is optional and we need to have at least one expected property.

PR #257, opened 2020-10-15
Improve the logic of required properties for article image

nlisgo:
Although this is failing tests, the change to the required parameter needs to be only made to the schema where the social image is expected. 

The behaviour should be:

1. the `image` property is optional.
2. in article listings if the image property is present it should only be for `vor` and must include the `thumbnail` property.
3. for `poa` full articles if the `image` property is there it must contain the `social` property.
4. for `vor` full articles if the `image` property is there it must contain either the `thumbnail` or `social` property or both.

Before this change 1, 3 and 4 was true but 2 wasn't.

Because the test suite doesn't currently have examples of expected errors it is hard to demonstrate that all of these conditions are true.

cc @lsh-0, I will put some more time on this tomorrow. I am having the same issue with the `interview` content on this PR: https://github.com/elifesciences/api-raml/pull/256

cc @thewilkybarkid @giorgiosironi any pointers would be really appreciated.

### DIFF
--- a/src/snippets/article-vor-basic.v1.yaml
+++ b/src/snippets/article-vor-basic.v1.yaml
@@ -1,0 +1,21 @@
+$schema: http://json-schema.org/draft-04/schema#
+title: Article VoR snippet (basic)
+type: object
+allOf:
+  - $ref: article.v1.yaml
+  - properties:
+        status:
+            type: string
+            enum:
+              - vor
+        impactStatement:
+            type: string
+            minLength: 1
+        figuresPdf:
+            type: string
+            format: uri
+        image:
+            type: object
+            properties:
+                thumbnail:
+                    $ref: ../misc/image.v1.yaml
--- a/src/common/article-vor.v1.yaml
+++ b/src/common/article-vor.v1.yaml
@@ -1,8 +1,8 @@
 $schema: http://json-schema.org/draft-04/schema#
-title: Article VoR snippet (basic)
+title: Article VoR common properties
 type: object
 allOf:
-  - $ref: article.v1.yaml
+  - $ref: ../snippets/article.v1.yaml
   - properties:
         status:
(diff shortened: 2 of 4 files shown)


lsh-0:
> cc @lsh-0, I will put some more time on this tomorrow.

that's fine, you're not blocking me on anything.

nlisgo:
I cannot seem to override the `required` property in `src/model/article-vor.v4.yaml` from `../snippets/article-vor.v1.yaml`. I don't love this solution but I think we should merge it in as it ensures the schema is correct. I will bring this PR up with @giorgiosironi in our next 1-to-1.

src/snippets/article-vor-basic.v1.yaml
```diff
@@ -0,0 +1,21 @@
...
+            type: string
+            minLength: 1
+        figuresPdf:
+            type: string
+            format: uri
+        image:
+            type: object
+            properties:
+                thumbnail:
+                    $ref: ../misc/image.v1.yaml
```
nlisgo: this contains everything from `src/snippets/article-vor.v1.yaml` but without the `required` property. If this is set in `src/snippets/article-vor.v1.yaml` I am finding it difficult to override in the full article view where `thumbnail` may not be present but either `thumbnail` and/or `social` should be expected.

NuclearRedeye:
> I cannot seem to override the required property

After spending sometime today looking into this, there _isn't_ anyway to do this with JSON schema by design. Hence this is in my opinion the correct approach. Move the common properties into a new file, and then simply declare the extras from image in the relevant snippet and model files. I felt that it would make more sense to move this common files into a new location to make it clearer that they are shared between various schemas, hence I pushed up the above adjustment. There are other files in the snippets directory that we should also move IMO, but not within the scope of this issue!

PR #258, opened 2020-10-20
Update maintainers

NuclearRedeye:
Need to update the maintainers of this project to reflect the new organisation structure. As the API specification is a rather critical keystone to a lot of the other projects it shouldn't have just a single maintainer, but rather multiple. Hence, I've raised this PR to start the discussion.
@gnott I'd like you to be on the list also, but thought I'd ask if you'd prefer your elifesciences email over your personal one?

### DIFF
--- a/maintainers.txt
+++ b/maintainers.txt
@@ -1,1 +1,3 @@
-c.wilkinson@elifesciences.org
+j.summerfield@elifesciences.org
+n.lisgo@elifesciences.org
+l.skibinski@elifesciences.org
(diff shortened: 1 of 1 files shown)


nlisgo (approved):

gnott:
I have not an elife email address (never have) and the one I have now is pretty much everywhere I would say, so it would be fine to include, if you wish to.

NuclearRedeye:
@gnott Understood, I'd always assumed that you had an elifesciences address but configured just as a forwarder!

lsh-0 (approved):

lsh-0:
> I'd always assumed that you had an elifesciences address but configured just as a forwarder!

Also, when sending invites to Google docs remember to add @gnott when setting the sharing/visibility rules.

PR #260, opened 2021-02-12
Add support for optional social image in digest

nlisgo:


lsh-0:
If I'm reading the changes correctly, a social-image now becomes required. Is all the digests content without a social image now invalid?

nlisgo:
> If I'm reading the changes correctly, a social-image now becomes required. Is all the digests content without a social image now invalid?

Social image is optional, there are plenty of samples of digest without a social image in this project and it is confirmed by the green test run here: https://github.com/elifesciences/api-dummy/pull/175

src/snippets/digest.v1.yaml
```diff
@@ -26,6 +26,8 @@ properties:
         properties:
             thumbnail:
                 $ref: ../misc/image.v1.yaml
+            social:
+                $ref: ../misc/image.v1.yaml
         required:
           - thumbnail
```
nlisgo, line 32: @lsh-0 only `thumbnail` is required.

lsh-0 (approved):

PR #261, opened 2021-02-19
Add support for references of type journal without page

nlisgo:


src/samples/article-vor/v5/complete.json
```diff
@@ -0,0 +1,1783 @@
...
+                {
+                    "type": "person",
+                    "name": {
+                        "preferred": "Person One",
+                        "index": "One, Person"
+                    }
+                }
+            ],
+            "articleTitle": "Minimal journal article",
+            "journal": "A journal"
```
nlisgo, line 1110: This is an example of a reference of type journal without pages

gnott (approved):
Looks good to me, thanks @nlisgo!

lsh-0:
Didn't get around to reviewing this today, @nlisgo , sorry. It's on my list for tomorrow as well as gauging some idea of the impact on lax.

lsh-0 (approved):
From my reading of the changes, this is what is happening:

- adds a *reference-v2*
    - only difference from *v1* is `page` is no longer a required property
- deprecates *article-vor-v4*
- adds *article-vor-v5*
    - *article-vor-v5* requires new *reference-v2* type references

Assuming my take is correct then the schema is being relaxed rather than made more strict with new constraints. Articles and references that were valid before will continue to be valid.

I don't think a major version bump is necessary but it can't hurt. It will require some minor code changes to lax but nothing serious.

lsh-0:
this is something else I really think we should consider, starting here: we keep a changelog for this repository.

https://keepachangelog.com/en/1.0.0/

Of all of our projects that would benefit the most from a changelog, it would be this one.

Breaking down my summary from above into changelog format:

```markdown
# Changelog
## 0.0.x
### Added

* a reference-v2, dropping the 'required' constraint on the 'page' property
* a article-vor-v5, that uses reference-v2 type references

### Changed

* article-vor-v4 is now *deprecated* and will be removed in a future release

### Fixed

*no fixes*

### Removed

*no removals*
```

fyi @NuclearRedeye 

update: I've opened a PR that will merge into this PR. If we can get a consensus that a changelog is a good idea and worth keeping for api-raml then just merge it in.

lsh-0:
I've added the proposed changelog.

Once merged it will:

1. go through testing on Jenkins
2. automatically update the `master` branch when tests pass
3. automatically compile the contents of `src` into `dist` and update the `master` branch.

This last commit needs to be tagged `1.1.0` manually.

    cd /path/to/api-raml
    git checkout master
    git pull
    git tag 1.1.0
    git push --tags

Github will create a '[release](https://github.com/elifesciences/api-raml/releases)' and, optionally, the description of this release should be updated with the changelog entry. Otherwise it will just be `Regenerated dist/`.

nlisgo:
> Assuming my take is correct then the schema is being relaxed rather than made more strict with new constraints. Articles and references that were valid before will continue to be valid.

Changing a property that was mandatory to be optional is indeed a relaxing of the rules but client code may be written to just access the expected value without checking that it exists first and therefore when it becomes optional the client code breaks. This is why I believe a version bump is necessary in this instance.

~@lsh-0 are you happy for this to be merged?~ I'm inferring from your last comment that you are.

nlisgo:
@gnott please merge in when you are ready to work on the implementation of these changes.

Feel free to create the tag `1.1.0` after the changes have been merged into master but I don't mind taking care of that afterwards.

gnott:
I tested some temporary changes to parser code and `bot-lax-adaptor` validation and things are looking good, however because the new `dist` files in this PR aren't generated yet, I had to fake it. I'll merge it and put together the other parts into a `bot-lax-adaptor` PR to review.

gnott:
Tagged `1.1.0`

PR #262, opened 2021-02-23
adds a changelog

lsh-0:
This will add a changelog to the PR https://github.com/elifesciences/api-raml/pull/261 where I suggested it.

*That* PR can then be merged into `develop` and then `git tag` to `0.1.0`.

Feel free to make any suggestions. It will require some effort from the developers making changes to api-raml going forwards so I don't want to force the change through. 

Approve the PR if you think it's a good idea.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 # Changelog
-## 0.1.0
+## 1.1.0
 ### Added
 
 * a reference-v2, dropping the 'required' constraint on the 'page' property
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,19 @@
+# Changelog
+## 0.1.0
+### Added
+
+* a reference-v2, dropping the 'required' constraint on the 'page' property
+* a article-vor-v5, that uses reference-v2 type references
+* this changelog!
+
+### Changed
+
+* article-vor-v4 is now *deprecated* and will be removed in a future release
+
+### Fixed
+
+*no fixes*
+
+### Removed
+
+*no removals*


CHANGELOG.md
```diff
@@ -0,0 +1,19 @@
+# Changelog
+## 0.1.0
```
nlisgo: Perhaps we should tag the current code in master as `1.0.0` and this would be `1.1.0`?

Also, would it make sense to adjust the Jenkins pipelines and fail ci if there are changes to the `dist` folder and no changes to the changelog?

Would version bumps (the git tag bit) be done by a jenkins pipeline after merged into master or manually?

NuclearRedeye (approved):
I'd be really interested in discussing this more, but for now this is a positive step forward and as such I'm happy for it to be committed as is.

NuclearRedeye:
> Also, would it make sense to adjust the Jenkins pipelines and fail ci if there are changes to the dist folder and no changes to the changelog?

Sounds like a good idea and something that we should followup on.

> Would version bumps (the git tag bit) be done by a jenkins pipeline after merged into master or manually?

Ideally, but it's difficult. When I first setup the Editor project, I used a combination of tools (standard-version, conventional-commits, husky) to ensure that commit messages followed a specific format and hence versioning was automatic based on the commit messages between versions. It worked well to be honest, but it wasn't without it's downsides and required devs to be disciplined in writing meaningful commit messages, something which (as pointed out in https://keepachangelog.com/en/1.0.0/) sadly doesn't really work out in reality.

gnott:
I wouldn't standing in the way of maintaining changelogs, as long as I remember there is one that requires updating (not that I would be adding to this particular one very often, but it's come up when we discsused `elife-tools` parser version releases).

How would the proposed API RAML version number relate to the https://packagist.org/packages/elife/api package and how versions are recorded there?

gnott (approved):
A changelog is better than no changelog! It would be cool if we leave space to potentially add historical major changes to the schema if it's useful, to list breaking changes, although it doesn't seem to have affected any non-eLife users in the past of this schema.

CHANGELOG.md
lsh-0: good call. I've just tagged `master` as `1.0.0` and changed this to `1.1.0`.

it's always fairly arbitrary when something hits the big 1.0, but you have to start somewhere and api-raml has been stable for ages.

lsh-0:
> Also, would it make sense to adjust the Jenkins pipelines and fail ci if there are changes to the dist folder and no changes to the changelog?

and

> Would version bumps (the git tag bit) be done by a jenkins pipeline after merged into master or manually?

This is an interesting point. Also, the version that is tagged should be the version where `dist` is regenerated.

For me, changelogs and release tagging seem to require a considerate human touch. It requires the author of the changelog to know their audience (other developers) and why they would be reading it in the first place. When I read a changelog I want to know what is going to break when upgrading, any new features that are available, bugs that were fixed that I didn't know about, deprecations, obsolescence, etc.

Automating the creation of a changelog, figuring out the right tag, pre-commit checks for changes to files etc is definitely possible with a lot of cleverness, but, long-term, cleverness always seems to be high-maintenance, inflexible and cultivates ignorance and a dependence on developer-magicians who know all the tricks. I'm not against automation at all - it's my very job! - but the 'semantic' in 'semantic versioning' means a human has to decide if the change is a major/minor/patch version and how best to communicate that to their audience, and speaking for myself, I don't have the time or patience to mentally parse out what the *actual* implications of a new release are from an automatic changelog. I want to be spoonfed that information.

This is a long winded way of saying that I think a changelog should be handcrafted and the correct git revision tagged manually.

I *do* think releasing software would benefit from a sort of 'staging' process where the commit messages since last release are marshaled together to form a draft changelog, a new version tag is *suggested*, an optional checklist of items *per-project* are presented, etc.

For now, getting developers to feel the rhythm of releasing software is more important than automating away the rough edges.

> I wouldn't standing in the way of maintaining changelogs, as long as I remember there is one that requires updating (not that I would be adding to this particular one very often, but it's come up when we discsused elife-tools parser version releases).

I think emphasis on versioning and maintaining a changelog is new for elife devs, something that didn't happen before because of the breakneck speed of development and generally figuring things out as we went along. Emphasis has shifted towards maintaining these projects now and strong versioning and changelogs and a slower pace overall are hallmarks of more stable software.

> How would the proposed API RAML version number relate to the https://packagist.org/packages/elife/api package and how versions are recorded there?

I just tagged `master` as `1.0.0` and if you hit that link now you can see `1.0.0` in the list. It seems to be tracking it fine but I don't know if it's automatic or if it triggered something in Jenkins ...

I also don't know much about packagist.org but was able to login just now as the `elife-projects-builder` user, the only credentials we have on file. The account has an email address set for failed builds but no packages configured. It looks unused to me. I think Chris is [using his own account](https://packagist.org/users/thewilkybarkid/) to publish php packages.

PR #263, opened 2021-03-01
Remove below version 4 vor and version 2 poa

nlisgo:
### Added

*no additions*

### Changed

*no changes*

### Fixed

* a couple of typos in api.raml

### Removed

* removed support for article-vor-v3
* removed schema model/article-poa.v1.yaml
* removed schemas model/article-vor.v1.yaml, model/article-vor.v2.yaml and model/article-vor.v3.yaml


gnott (approved):
Looks good to me!

CHANGELOG.md
```diff
@@ -1,4 +1,23 @@
 # Changelog
+## 1.2.0
```
lsh-0: we're removing things so this needs to be `2.0.0` rather than `1.2.0`

CHANGELOG.md
```diff
@@ -1,4 +1,23 @@
...
+### Added
+
+*no additions*
+
+### Changed
+
+*no changes*
```
lsh-0, line 9: When a section is empty we can probably just leave it out altogether. Up to you.

CHANGELOG.md
lsh-0: for example, if I were using composer or similar to describe dependencies for a project that used api-raml and I had specified  `~1.1` then I'd be surprised that when the lockfile was updated and the upgrade to `1.2` was made that article vor v3 support was removed. The project depended on it but had been too lazy (or busy ;) to handle any deprecation warnings.

`~1.1` wouldn't upgrade to `2.0` however.

lsh-0 (approved):

PR #264, opened 2021-03-02
Support image thumbnail in article poa snippet and full item

nlisgo:


CHANGELOG.md
```diff
@@ -1,4 +1,23 @@
 # Changelog
+
+## 1.2.0
```
lsh-0, line 3: we need to be careful about the order of merging now. 

this is 1.2.0 and the old vor removal is 2.0.0. this will have to come first.

lsh-0 (approved):

PR #265, opened 2021-03-08
Alter samples of preferred name for reference and dataset author lists

nlisgo:


CHANGELOG.md
```diff
@@ -1,4 +1,21 @@
 # Changelog
+## 2.0.1
+### Added
+
+*no additions*
+
+### Changed
+
+* altered the sample values for author preferred names in references and datasets
```
lsh-0: Not a good description. Perhaps:

> author `preferred` field changed from "surname firstname" to "firstname surname" in references and datasets

CHANGELOG.md
```diff
@@ -1,4 +1,21 @@
 # Changelog
+## 2.0.1
```
lsh-0: Patch versions are for backwards compatible bugfixes typically (https://semver.org).

In this case we're probably going for `2.1.0`. It's still a significant change but it doesn't break our public API.

CHANGELOG.md
lsh-0: looking at the article changes I guess the "One Person" and "Person One" examples aren't that clear. If this is what is happening:

>\-                         "preferred": "C Darwin",
>\+                       "preferred": "Darwin C",

Then the description would be:

> author preferred field changed from "firstname surname" to "surname firstname" in references and datasets

Anyway, your decision.

lsh-0 (changes requested):
that is a helluva lot of changes to have made, I hope you wrote a script :| 

I didn't see any other change except to examples, so I guess the actual logic lives in the xml/bot-lax.

apart from the description and version number in the changelog it should be good to go.

CHANGELOG.md
lsh-0: I think arguments could be made for this being a major or minor *or* patch version the more I think about it. Minor version still seems more sensible to me, preserving the patch version for more innocuous changes later.

nlisgo:
> that is a helluva lot of changes to have made, I hope you wrote a script :|

Regex!

> I didn't see any other change except to examples, so I guess the actual logic lives in the xml/bot-lax.

Yes.

> apart from the description and version number in the changelog it should be good to go.

Changed to minor version bump and amended the description. Found some more examples in the samples that I've added to the PR. Approve if you're happy and I'll merge it in on Thursday.

lsh-0 (approved):

PR #267, opened 2021-03-09
chore: modify the maintainers.txt file

nlisgo:


CHANGELOG.md
```diff
@@ -1,4 +1,21 @@
 # Changelog
+## 2.0.1
```
NuclearRedeye: Don't myself think we'd need to bump the version number for this sort of change as it doesn't effect the API.

lsh-0: Yes, not every change gets mentioned in the changelog.

@nlisgo , imagine you're looking at upgrading an important library to a new version. You want to make sure you have a rough understanding of what has changed between the version you're currently using and what you'll be upgrading to. Have any bugs been fixed? New features added? Stuff you relied upon changed significantly or removed? The changelog is a simple list of that. 

For really fine-grained information the commit log could be consulted.

Changes can accumulate before we decide to do a proper release.

PR #269, opened 2021-06-15
article history, preprint events.

lsh-0:
attempt 3

### DIFF
--- a/src/api.raml
+++ b/src/api.raml
@@ -510,7 +510,16 @@ traits:
                                         description: |
                                             Multiple versions including a version being previewed.
                                         value: !include samples/article-history/v1/unpublished-multiple-versions.json
-
+                                    preprintOnly:
+                                        displayName: Pre-print only
+                                        description: |
+                                            Article history involving just a pre-print. Valid, but unlikely to ever exist.
+                                        value: !include samples/article-history/v1/preprint-only.json
+                                    oneVersionWithPreprint:
+                                        displayName: One version and a pre-print
+                                        description: |
+                                            Article history involving a published article and a pre-print.
+                                        value: !include samples/article-history/v1/one-version-with-preprint.json
             /{version}:
                 type:
                     item:
--- a/src/api.raml
+++ b/src/api.raml
@@ -2,7 +2,7 @@
 ---
 title: eLife Sciences API
 baseUri: https://api.elifesciences.org/
-version: 1.0.0
+version: 2.2.0
 documentation:
   - title: Welcome
     content: |
@@ -510,16 +510,19 @@ traits:
                                         description: |
                                             Multiple versions including a version being previewed.
                                         value: !include samples/article-history/v1/unpublished-multiple-versions.json
+                            application/vnd.elife.article-history+json;version=2:
+                                schema: !include ../dist/model/article-history.v2.json
+                                examples:
                                     preprintOnly:
                                         displayName: Pre-print only
                                         description: |
                                             Article history involving just a pre-print. Valid, but unlikely to ever exist.
-                                        value: !include samples/article-history/v1/preprint-only.json
+                                        value: !include samples/article-history/v2/preprint-only.json
                                     oneVersionWithPreprint:
                                         displayName: One version and a pre-print
                                         description: |
                                             Article history involving a published article and a pre-print.
-                                        value: !include samples/article-history/v1/one-version-with-preprint.json
+                                        value: !include samples/article-history/v2/one-version-with-preprint.json
             /{version}:
                 type:
                     item:
--- a/src/samples/article-history/v1/one-version-with-preprint.json
+++ b/src/samples/article-history/v1/one-version-with-preprint.json
@@ -1,0 +1,27 @@
+{
+    "received": "2016-01-02",
+    "accepted": "2016-02-28",
+    "versions": [
+        {
+            "status": "vor",
+            "id": "14107",
+            "version": 1,
+            "type": "research-article",
+            "doi": "10.7554/eLife.14107",
+            "authorLine": "Yongjian Huang et al.",
+            "title": "Molecular basis for multimerization in the activation of the epidermal growth factor",
+            "stage": "published",
+            "published": "2016-03-28T00:00:00Z",
+            "versionDate": "2016-03-28T00:00:00Z",
+            "statusDate": "2016-03-28T00:00:00Z",
+            "volume": 5,
+            "elocationId": "e14107"
+        },
+        {
+            "type": "preprint",
+            "description": "This manuscript was published as a preprint at bioRxiv.",
+            "uri": "https://www.biorxiv.org/content/10.1101/2019.08.22.6666666v1",
+            "date": "2010-01-01T00:00:00Z"
+        }
+    ]
+}
--- a/src/samples/article-history/v1/preprint-only.json
+++ b/src/samples/article-history/v1/preprint-only.json
@@ -1,0 +1,10 @@
+{
+    "versions": [
+        {
+            "type": "preprint",
+            "description": "This manuscript was published as a preprint at bioRxiv.",
+            "uri": "https://www.biorxiv.org/content/10.1101/2019.08.22.6666666v1",
+            "date": "2019-02-15T00:00:00Z"
+        }
+    ]
+}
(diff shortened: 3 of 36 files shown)


lsh-0:
OK, third times the charm. With examples this time.

This approach makes more sense than the previous two imo.

src/api.raml
```diff
@@ -510,7 +510,16 @@ traits:
                                         description: |
                                             Multiple versions including a version being previewed.
                                         value: !include samples/article-history/v1/unpublished-multiple-versions.json
-
+                                    preprintOnly:
```
lsh-0, line 516: @nlisgo , do you think the inclusion of preprints should bump `article-history` from `v1` to `v2` ?

lsh-0: if we bump this to `v2` there might be a bit more code/coordination required, but I could exclude preprint data from lax results so those requesting `v1` article-history data will continue working as normal and never receive preprints.

src/snippets/article-preprint.v1.yaml
```diff
@@ -0,0 +1,21 @@
+$schema: http://json-schema.org/draft-04/schema#
+title: Article pre-print version snippet
+type: object
+properties:
+    type:
```
gnott: My observation is although `type` probably matches the XML terminology, the value of "preprint" is similar to the `status` values like `poa` and `vor`. The snippet definitions can be like apples and oranges, and I don't feel strongly either way.

src/samples/article-history/v1/one-version-with-preprint.json
```diff
@@ -0,0 +1,27 @@
...
+            "published": "2016-03-28T00:00:00Z",
+            "versionDate": "2016-03-28T00:00:00Z",
+            "statusDate": "2016-03-28T00:00:00Z",
+            "volume": 5,
+            "elocationId": "e14107"
+        },
+        {
+            "type": "preprint",
+            "description": "This manuscript was published as a preprint at bioRxiv.",
+            "uri": "https://www.biorxiv.org/content/10.1101/2019.08.22.6666666v1",
```
giorgiosironi: note this is a version-specific URL, and bioRxiv preprints can have multiple versions too (which this API is probably not interested in capturing)

src/samples/article-history/v1/preprint-only.json
```diff
@@ -0,0 +1,10 @@
+{
+    "versions": [
+        {
```
nlisgo: Currently, each of the versions of an article can be accessed at the endpoint `/articles/[article-id]/version/[version-number]`. Introducing versions without version numbers seems wrong.

I suggest that we add another optional top level property of `preprint` with the required properties of `uri` and `date` and optional `description` property.

src/samples/article-history/v1/one-version-with-preprint.json
lsh-0: it's a fair point and something @FAtherden-eLife might have thought about already. As far as the spec and elife-tools are concerned we're just using whatever we pull out of the XML.

src/samples/article-history/v1/preprint-only.json
lsh-0: > Currently, each of the versions of an article can be accessed at the endpoint /articles/[article-id]/version/[version-number]. Introducing versions without version numbers seems wrong.

It's not a known version, you won't be able to access it at `/articles/[article-id]/version/[version-number]`

> I suggest that we add another optional top level property of preprint with the required properties of uri and date and optional description property.

Do you mean like:

```json
{
    "preprint": "2015-12-15",
    "preprint-description": "...",
    "received": "2016-01-02",
    "accepted": "2016-02-28",
    "versions": [
          ...
    ]
}
```

It's possible for an article to have multiple preprints so that won't work. [Discussion ticket here](https://github.com/elifesciences/issues/issues/4702). 

This change is modelling an 'event' in the article's history. There are other event types planned (edit: [here](https://github.com/elifesciences/issues/issues/4702#issuecomment-853699910-permalink)), for now we're capturing the publication a preprint.

src/snippets/article-preprint.v1.yaml
lsh-0: From [this comment](https://github.com/elifesciences/issues/issues/4702#issuecomment-853699910-permalink) we can expect these as well:

* preprint
* accepted-manuscript
* version-of-record
* corrected-version-of-record

... although, we already have a version-of-record (`status: VOR`) 'event' and I guess a 'corrected-version-of-record' would simply be a VOR v2, v3, etc. And accepted-manuscript is POA. So ... I guess we can ignore those when they come along and @nlisgo's suggestion about making the preprint data top-level has more merit. 

I still think that now that there is an extension point in the XML it will be used. Just like we have articles with 5 or 6 VOR versions, I wouldn't be surprised to see an article with multiple preprint events.

lsh-0:
The feedback has been good guys, thanks. 

I'm going to:
* bump the article-history version endpoint because introducing this preprint history will probably break apps. PHP apps requesting article-history v1  will still continue to work and will need to opt-in to receive preprints.
* change preprint `type` to `status` for consistency

src/api.raml
lsh-0: I went ahead and bumped it. I think a rule of thumb when deciding to increment a spec version is "does it break things?" if yes, bump.

src/snippets/article-preprint.v1.yaml
lsh-0: changed `type` to `status` for consistency.

lsh-0:
OK, changes made. 

I've also made the `version` attribute in the `api.raml` file consistent with the value in the changelog.

lsh-0:
I'd like to merge this tomorrow so please let me know if you think there are design flaws or deal breakers in there.

src/samples/article-history/v1/preprint-only.json
nlisgo: I was thinking of a single extra property of `preprint` which was an object but it could also be an array. It was really because it felt uncomfortable to introduce "versions" that would not be accessible at the `/articles/[article-id]/versions/[version-id]` endpoint.

In practice will we be updating our XML each time that a preprint version has been deposited on bioRxiv? Or will we only be referring to the latest preprint?

lsh-0: it wouldn't be a pre-print then but a post-print ...

nlisgo:
I will approve this. I may have some more feedback as I implement this that can feed into the next version. Great work @lsh-0

nlisgo (approved):

src/samples/article-history/v1/preprint-only.json
nlisgo: > it wouldn't be a pre-print then but a post-print

Fair point.

Another question then. In practice will we be listing all versions of preprint that have been deposited or only the latest one?

lsh-0: It will be possible for an article to have multiple preprints attached to it, I assume from different pre-print servers. I've not read anything about just a single pre-print being displayed when multiple are available. I would assume all would be used but feel free to bring it up in the discussion: https://github.com/elifesciences/issues/issues/4702

src/samples/article-history/v1/one-version-with-preprint.json
fred-atherden: > (which this API is probably not interested in capturing)

Yes, we're not interested in capturing the different versions.

> note this is a version-specific URL

We may use the DOI uri instead (i.e. `https://doi.org/10.1101/2019.08.22`) which will resolve to the latest version. Since it'll still be a resolvable uri, presumably this won't cause any problems here?

lsh-0: nah, no problems. I'll update the example so it's less confusing.

lsh-0:
There was some good feedback to questions in the discussion ticket: https://github.com/elifesciences/issues/issues/4702

Multiple pre-prints weren't ruled out but it seems they would be unlikely. Not as close to "definitely only one preprint" as I would  like to swing the design towards @nlisgo suggestions. 

With no XML spec presented to enforce single preprints and the possibility of future event types other than the 'preprint' that I can't anticipate right now I'm going to err on the side of caution and go with the current design.

Lists of things are an easy and natural type of extension point, unfortunately they won't be a list of homogenous types anymore with a full version accessible via the `/articles/{msid}/versions/{version}` endpoint.

Merging.

PR #270, opened 2021-07-01
duplicate "application/vnd.elife.article-history+json; version=1" examples in "version=2"

nlisgo:
Also mark version 1 as deprecated.

Have add a changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 2.3.0
+## 2.2.1
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 2.3.0
+
+### Changed
+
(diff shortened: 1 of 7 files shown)


lsh-0:
Should probably be ~`2.3.1`~ `2.2.1`, we're 'patching' something I should have added in ~`2.3.0`~ `2.2.0`. semver doesn't directly apply to non-code so I'm still going by gut instinct here. It's still a good thing to do I think even if we quibble over what constitutes a major/minor/patch version.

I updated this value last time as well: https://github.com/elifesciences/api-raml/blob/develop/dist/api.raml#L5

lsh-0:
edit: 

> Should probably be `2.3.1`

I meant, `2.2.1`

nlisgo:
> I meant, `2.2.1`

If only for adding the samples I would agree but marking the previous media type version as deprecated felt like more of a minor change. But, I'm happy to change to `2.2.1` as suggested. Will add a commit.

nlisgo:
@lsh-0 please approve or add additional feedback. I believe this is ready to roll.

lsh-0 (approved):

PR #271, opened 2021-07-28
Editor evaluation

gnott:
Re issue https://github.com/elifesciences/issues/issues/6758

To start with a first draft of adding `editorEvaluation` content. Comments welcome please!

PR #272, opened 2021-07-28
Editor evaluation

gnott:
Re issue https://github.com/elifesciences/issues/issues/6758

Re-do of closed PR https://github.com/elifesciences/api-raml/pull/271

A first draft of adding `editorEvaluation` content. Comments welcome please!

---

An example of the XML in question. 

```xml
<article article-type="research-article" dtd-version="1.2" xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:ali="http://www.niso.org/schemas/ali/1.0/">
    ...
    <sub-article article-type="editor-report" id="sa0">
        <front-stub>
            <article-id pub-id-type="doi">10.7554/eLife.62927.sa0</article-id>
            <title-group>
                <article-title>Editor's evaluation</article-title>
            </title-group>
            <contrib-group>
                <contrib contrib-type="author">
                    <name>
                        <surname>Ma</surname>
                        <given-names>Yuting</given-names>
                    </name>
                    <role specific-use="editor">Reviewing Editor</role>
                    <aff>
                        <institution>Suzhou Institute of Systems Medicine</institution>
                        <country>China</country>
                    </aff>
                </contrib>
            </contrib-group>
        <related-object link-type="hasRelatedMaterial" xlink:href="https://sciety.org/articles/activity/10.1101/2020.11.21.391326"/>
        </front-stub>
        <body>
            <p>Collagen is a major component of extracellular matrix. The authors have identified a high-affinity inhibitory collagen receptor LAIR-1 and a soluble decoy receptor LAIR-2 (with even higher binding affinity to collagen), which can be therapeutically targeted to block tumor progression. Dr Meyaard and colleagues have also generated a dimeric LAIR-2 human IgG1 Fc fusion protein NC410 for therapeutic use. With humanized mouse models engrafted with functional human immune systems (PBMC), they have explored the anti-cancer efficacy of NC410 and revealed its impact on modulating immune responses. Furthermore, they extended this study to identify biomarkers of predictive value for NC410-based anti-cancer therapy.</p>
        </body>
    </sub-article>
    ...
</article>
```


lsh-0:
I'll take a look today

src/model/article-vor.v6.yaml
```diff
@@ -65,6 +65,24 @@ allOf:
             items:
                 $ref: ../misc/blocks-all.v2.yaml
             minItems: 1
+        editorEvaluation:
+            type: object
+            properties:
+                id:
+                    $ref: ../misc/html-id.v1.yaml
+                doi:
+                    $ref: ../misc/doi.v1.yaml
```
lsh-0, line 74: looks like: `10.7554/eLife.62927.sa0`

@gnott I remember years ago a decision to stop using sub-dois - was that limited to just abstracts or figures or something similar?

lsh-0:
author information appears to be missing

lsh-0:
I'll hold off on any decision until it's more complete. Feel free to re-request closer to completion.

src/model/article-vor.v6.yaml
gnott: You remember correctly there was a change years ago, specifically the sub-dois were being reported to Crossref as component DOIs. What remained for more recent articles was a sub-dois for each of the `<sub-article>` XML tags -- the decision letter, and author response, and now is to include an editor's evaluation -- and these are reported to Crossref as peer review materials. The unique DOI allows each of them to be registered as such and can link back to the main article's DOI.

gnott:
> author information appears to be missing

I asked one short question on a call about the contributor names, and the conclusion was they were likely not going to be important to include in the API JSON output, much like for the decision letter it does not include the specific contributor / editor names included in the article XML. For this reason I didn't try to define them here and I followed the format of the existing peer review articles here.

I think I do end up parsing the contributors to make Crossref deposits, which uses the article XML as the input. I would guess places like PMC also do something with the data in the article XML.

lsh-0 (approved):

src/samples/article-vor/v6/complete.json
```diff
@@ -1107,8 +1107,7 @@
                 }
             ],
             "articleTitle": "Minimal journal article",
-            "journal": "A journal",
-            "pages": "In press"
```
nlisgo, line 1111: Was the `pages` property removed in error?

src/model/article-vor.v6.yaml
```diff
@@ -101,5 +119,3 @@ allOf:
               - content
     required:
       - body
-    dependencies:
-        authorResponse: [decisionLetter]
```
nlisgo, line 105: If I understand this `dependencies` block correctly it means that where there is an `authorResponse` property set we should require a `decisionLetter` to be set?

Should we be removing this entirely or do we need to express a more complex `dependencies` block which states that if an `authorResponse` is present would should expect one of `decisionLetter` or `editorEvaluation`?

gnott: I looked briefly at JSON schema to see how you might be able to specify a "this or that" structure and I didn't find anything quickly. If you can figure out how to satisfy the the complexity it would be welcome I think. Do you anticipate the removal of the dependencies to cause any programming challenges if we had to remove it?

src/samples/article-vor/v6/complete.json
gnott: There's some weird display issue here because I did both a `git rm src/samples/article-vor/v4/complete.json` and I added `/v6/complete.json` which is an exact copy of `/v5/complete.json`, this is wrongly concluding the v4 has been turned into v6. Therefore, if this data was removed, it was at the time the `/v5/complete.json` file was added.

nlisgo: That makes sense. I consider this resolved.

src/model/article-vor.v6.yaml
nlisgo: > Do you anticipate the removal of the dependencies to cause any programming challenges if we had to remove it?

No, but it's about reflecting in the schema, as much as possible, the expectations of the content properties. I don't want to lose what we already have in place here. Currently, if we have JSON that includes an `authorResponse` but not a `decisionLetter` then this issue would be flagged and we could fix it. I will try and work on this change today and we could add a couple more samples to prove this.

nlisgo:
@gnott I've improved the dependencies to expect either a `decisionLetter` or `editorEvaluation` if `authorResponse` is present.

nlisgo:
I'm approving this but we will need to bump the version and add an entry to `CHANGELOG.md` before merging in.

nlisgo (approved):

gnott:
Very cool use of the `anyOf`, thanks @nlisgo,

gnott:
I added a commit bumping to version `2.3.0` and added some content to the CHANGELOG, I'll leave it for any comments about those bits. I tested validation for when it is only `editorEvaluation` and `authorResponse` and it worked properly when I tried validating using `bot-lax-adaptor`.

src/model/article-vor.v6.yaml
```diff
@@ -38,7 +38,7 @@ allOf:
         references:
             type: array
             items:
-                $ref: ../misc/reference.v1.yaml
+                $ref: ../misc/reference.v2.yaml
```
nlisgo, line 41: I think we can now remove the file `/src/misc/reference.v1.yaml` now

gnott: Thanks @nlisgo, I guess mentioning a file like this was removed in the CHANGELOG is a little too granular to bother putting there do you think? I don't find any other removal message there about any files in `/src/misc/` so I won't add it to the removal list at this time.

gnott:
I think this PR is ready to merge, I wanted to leave it here until @lsh-0 had a chance to review or comment again. I also think the next step might be in `bot-lax-adaptor` and/or `lax` and I'm not sure the extent of changes required there to support the new vor v6 version and the deprecation of vor v4

lsh-0:
yup, ready to merge.

>  I also think the next step might be ...

I create these coordination tickets so we don't have to keep guessing what the next steps are:

* pre-prints: https://github.com/elifesciences/issues/issues/6732
* journal references with no pages: https://github.com/elifesciences/issues/issues/6498

If we don't have one for the editor evaluation change, we should create one. Unless they're not helpful? I think they are. I'll find/create one.

gnott:
Yeah for sure, the co-ordination issue would be good I think if it doesn't already exist, please do create one. I need to create an issue for `elife-tools` to parse the new XML into `editorEvaluation` JSON too.

nlisgo:
We need to push the need tag `2.3.0`.

gnott:
I haven't found any existing pipelines that create the git tag, and I cannot find any mention of whether we tag it manually, so I'm going to create the tag now.

gnott:
Except that `elife-alfred-user` created the previous tags, I don't know where that happens, anyway tag exists now.

gnott:
Hmm, I think the tag was created on the `master` branch, and it isn't on the `develop` branch. I'll try to make sure things are all up to date.

nlisgo:
The tag seems to have taken. Thanks a lot @gnott!

PR #273, opened 2021-09-27
vor v7, uri changed to scietyUri in editorEvaluation.

gnott:
Re issue https://github.com/elifesciences/issues/issues/6940

Changes the `uri` property name to be `scietyUri`, which leads to establishing the next content type version `application/vnd.elife.article-vor+json;version=7`, adding samples for the vor v7 content, deprecating the vor v6 content type, removing the vor v5 content, and bumping this library's version number to `2.4.0`.

Please feel welcome to review these proposed changes and commit additional changes to this branch of code if I missed something or if there's an error.


lsh-0:
wait - have we even finished migrating off of v5 yet? https://github.com/elifesciences/issues/issues/6830

Doesn't look like it.

I don't see a coordination ticket/epic for this either.

Not approving anything until the current situation is clearer.

cc @NuclearRedeye

gnott:
Good catch @lsh-0 thanks. I guess maybe v5 to v6 transiiton would be held up if the `uri` value name is not liked. That might leave a couple options, either we leave three versions supported (v5, v6, and v7), or we just change `uri` to `scietyUri` in the v6, though that would go against version release conventions. I'll wait for others to weigh in, since this might be too early.

lsh-0:
I'm not pointing a finger at you Graham, but this situation is why I want the changes to the XML to be rock solid before we have to support them downstream. I think Fred (XML) and Nathan (api-raml) should put their heads together and come up with a good model first, get some feedback from the rest of us and *then* we commit ourselves. It may not prevent another situation like this but it should reduce them.

So we're stuck with a partial `v5 -> v6` migration that should be aborted because we have no intention of supporting `uri` because it's actually `scietyUri` we want to support.

I suggest we make this change part of the `v5 -> v6` migration and backfill any affected articles.

NuclearRedeye:
I thought this was a tweak/follow up to the previous changes to the API to add support for the Editors evaluation, hence when I spoke to @gnott yesterday I suggested he take a look hence the PR. I didn't realise that it might require a completely new version of the API, which if so we definitely don't want to have overlapping changes in progress.

These changes do come under the Editors Evaluation project, so if we can as @lsh-0 suggests include them as part of the `v5 to v6` migration that is already in progress then that sounds to me like a path forward. Otherwise it seems best to park this until all the current work is live and then follow up with this change after.

nlisgo:
How about this approach. As `scietyUri` is a new field we could introduce it as an optional field and just serve the same value as `uri`. Since `uri` is optional we risk nothing by removing it and we know it isn't being used anywhere yet. 

> I suggest we make this change part of the v5 -> v6 migration and backfill any affected articles.

Think this satisfies the above.

So, let's leave v5 in place and just progress v6.

lsh-0:
> and we know it isn't being used anyway anyone.

just to confirm, no articles are using an editor's evaluation yet?

nlisgo:
> just to confirm, no articles are using an editor's evaluation yet?

Not yet.

lsh-0:
should make things easier, cheers @nlisgo .

gnott:
Closing in favour of https://github.com/elifesciences/api-raml/pull/275

PR #275, opened 2021-10-04
Rename uri to scietyUri in editorEvaluation.

gnott:
Replaces PR https://github.com/elifesciences/api-raml/pull/273

Re issue https://github.com/elifesciences/issues/issues/6940

For review and further discussion, if required, I think this is the simpler method to change an existing property to a new name, `uri` to `scietyUri`, without incrementing the `article-vor` version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 2.5.0
+
+### Changed
+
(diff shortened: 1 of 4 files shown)


lsh-0 (approved):
yup, this seems fine to me if @nlisgo is ok with it as well.

nlisgo (approved):

PR #277, opened 2022-08-02
Add reviewed-preprint to the search v2 endpoint

smoqadam:


nlisgo:
@giorgiosironi will ci only be triggered for PR's that originate on this repo?

nlisgo:
I will discuss with @FAtherden-eLife whether we can drop some of the values such as volume and eLocationId.

This is marked as experimental so we don't need to worry about the superflous values. So, I think we should merge in once green and then address feedback in a followup.

PR #278, opened 2022-08-03
Add reviewed-preprint to the search v2 endpoint

nlisgo:


src/samples/search/v2/first-page.json
```diff
@@ -0,0 +1,250 @@
+{
+    "total": 13,
+    "items": [
+        {
+            "type": "reviewed-preprint",
+            "status": "reviewed",
+            "id": "09560",
```
nlisgo: We should ensure that this is unique

smoqadam: I opened another PR and changed the ID.

PR #282, opened 2022-08-09
Feat/reviewed preprint

smoqadam:
- add /reviewed-preprint endpoint
- add reviewed-preprint to /search


src/samples/reviewed-preprint/v1/complete.json
```diff
@@ -0,0 +1,43 @@
+{
+  "type": "reviewed-preprint",
```
nlisgo, line 2: We should add indexContent

PR #289, opened 2022-10-26
Add impact statement to cover's list

smoqadam:


nlisgo (approved):
Once this is merged in and the `/dist` folder is automatically regenerated we can create and publish the new tag `2.11.0`

PR #290, opened 2022-11-07
Add new samples for interview, podcast, collection, and POA article in covers-list

smoqadam:
…n covers-list

### DIFF
--- a/src/samples/cover-list/v1/first-page.json
+++ b/src/samples/cover-list/v1/first-page.json
@@ -88,6 +88,275 @@
                     }
                 }
             }
+        },
+        {
+            "title": "<i>Homo naledi</i>",
+            "impactStatement": "A new hominin species has been unearthed in the Dinaledi Chamber of the Rising Star cave system in the largest assemblage of a single species of hominins yet discovered in Africa.",
+            "image": {
+                "uri": "https://iiif.elifesciences.org/lax/09560%2Felife-09560-fig1-v1.tif",
+                "alt": "",
+                "source": {
+                    "mediaType": "image/jpeg",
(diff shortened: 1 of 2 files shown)


src/samples/cover-list/v1/first-page.json
```diff
@@ -88,6 +88,278 @@
...
+                },
+                "size": {
+                    "width": 4194,
+                    "height": 4714
+                }
+            },
+            "item": {
+                "type": "interview",
+                "id": "2",
+                "interviewee": {
```
nlisgo, line 163: The item only needs to have properties expected in a snippet. I can help with this.

src/samples/cover-list/v1/first-page.json
```diff
@@ -88,6 +88,278 @@
...
+                        "authorLine": "Yongjian Huang et al.",
+                        "title": "Molecular basis for multimerization in the activation of the epidermal growth factor",
+                        "stage": "published",
+                        "published": "2016-03-28T00:00:00Z",
+                        "statusDate": "2016-03-28T00:00:00Z",
+                        "volume": 5,
+                        "elocationId": "e14107"
+                    }
+                ]
+            }
```
nlisgo, line 284: We should add a blog-article if there isn't one already also. A few examples of properties that should not be present in the item as it is just the content snippet.

nlisgo (changes requested):

src/samples/cover-list/v1/first-page.json
smoqadam: The blog article is already there. That's why I didn't add it

nlisgo (approved):

PR #295, opened 2023-01-25
do not merge - regenerated dist for vor v7 with dependency constraints

nlisgo:
Have also fixed typo `recommendationForAuthors` to `recommendationsForAuthors`

This can be used instead of https://github.com/elifesciences/api-raml/pull/294


nlisgo:
@lsh-0 I just created this because I didn't want to mess with your PR since your were progressing with it.

lsh-0:
thanks @nlisgo , just some acrobatics we have to do when api-raml changes aren't in mainline. I'll repoint the lax and bot-lax work at this

PR #296, opened 2023-02-02
deprecate article-vor v6

nlisgo:
Now that vor v7 is live we should deprecate v6.


gnott (approved):
Clean and refreshing!

lsh-0 (approved):

PR #297, opened 2023-05-03
Add optional sentForReview date to article history

nlisgo:
@fred-atherden this is what we discussed yesterday. If you could update this PR with direction of where to find the value in the XML.

I'm not precious about what we call the field name but unless a better name surfaces let's run with this one.#

cc @lsh-0 @gnott @smoqadam

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 2.17.0
+
+### Added
+
+* introduce optional `sentForReview` date to article history
+
 ## 2.16.0
 
 ### Changed
--- a/src/model/article-history.v2.yaml
+++ b/src/model/article-history.v2.yaml
@@ -8,6 +8,9 @@ properties:
     accepted:
         type: string
         format: date
+    sentForPeerReview:
+        type: string
+        format: date
     versions:
         description: List of article versions, sorted by version in descending order
         type: array
--- a/src/model/article-history.v2.yaml
+++ b/src/model/article-history.v2.yaml
@@ -8,7 +8,7 @@ properties:
     accepted:
         type: string
         format: date
-    sentForPeerReview:
+    sentForReview:
         type: string
         format: date
     versions:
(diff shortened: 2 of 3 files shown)


fred-atherden (approved):

fred-atherden:
Thanks @nlisgo, In the XML the sent for review date can be found in the same place where accepted and received dates are in non-PRC content. Here it is in the kitchen sink:
https://github.com/elifesciences/XML-mapping/blob/f2dea9d0d0dd620219722aa31fa74189ea4079e2/elife-1234567890-v2/elife-1234567890-v2.xml#L189-L195

gnott (approved):
Looks simple enough and name is good.

nlisgo:
@fred-atherden can you confirm that there will only ever be one value for this? In the instance that a manuscript is a revised preprint could there possibly be multiple dates here?

fred-atherden:
@nlisgo, yeah - as far as I know we are only planning on showing one sent for review date. This date under PRC will be sort of equivalent to the accepted date under old model papers, which is why we don't intend to show later review dates in the timeline.

lsh-0 (approved):
Looks fine.

Can I get a link to the issue tracking this work?

gnott:
This is a related issue I think https://github.com/elifesciences/issues/issues/8172, except it may need updating or to be closed in favour of a more recent comprehensive epic issue.

nlisgo:
Thanks for the approval. I will create the epic today and you can add your tasks to it.

lsh-0:
How we going with this?

nlisgo:
Here is the epic: https://github.com/elifesciences/issues/issues/8311

Feel free to add tickets to it.

I propose that we change the property to `sentForReview` rather than `sentForPeerReview` to match the value in the XML.

nlisgo:
Updated to `sentForReview`.

nlisgo:
Expectation is that the date should be in the format: `YYYY-MM-DD`.

lsh-0:
Still fine, good to go.

nlisgo:
Added CHANELOG entry. Once it is green I will hit the merge button and add the tag once the `dist/` folder has been regenerated.

PR #298, opened 2023-05-18
Add support for reviewed-preprints in press-package

nlisgo:


src/samples/press-package/v4/complete.json
```diff
@@ -36,6 +55,52 @@
...
+            "type": "reviewed-preprint",
+            "doi": "10.7554/eLife.19560",
+            "status": "reviewed",
+            "authorLine": "Lee R Berger et al.",
+            "title": "<i>Homo naledi</i>, a new species of the genus <i>Homo</i> from the Dinaledi Chamber, South Africa",
+            "indexContent": "<i>Homo naledi</i>, a new species of the genus <i>Homo</i> from the Dinaledi Chamber, South Africa",
+            "titlePrefix": "Title prefix",
+            "stage": "published",
+            "published": "2022-08-01T00:00:00Z",
+            "reviewedDate": "2022-08-01T00:00:00Z",
```
gnott, line 71: If this is a `complete.json` file, would it be good to include a `versionDate` here too?

gnott (approved):
I'm not too familiar with these particular data models, they look good though and I have no additional feedback.

CHANGELOG.md
```diff
@@ -1,5 +1,20 @@
...
 
+## 2.17.0
+
+### Added
+
+* add experimental support for reviewed preprints in press packages in `application/vnd.elife.press-package+json;version=4`
+
+### Changed
+
+* removed experimental flag from `application/vnd.elife.press-package+json;version=3`
```
lsh-0: this would go under 'removed'

CHANGELOG.md
```diff
@@ -1,5 +1,20 @@
...
+
+### Removed
+
+* removed support for `application/vnd.elife.press-package+json;version=1`
+* removed support for `application/vnd.elife.press-package+json;version=2`
```
lsh-0: It's unusual to remove support for both previous versions (1, 2) and introduce two new versions (3, 4). There is no room to transition for services that depend on those content type versions.

I'd rather we support three versions (2, 3, 4) and mark the older two as deprecated (2, 3), following up with a separate PR to remove the oldest (2) after the downstream updates are complete.

I'd say that if we could absolutely guarantee nothing is using press-packages v1 or v2 then go for it, but obviously `journal` and `api-sdk-php` at the very least are using it.

lsh-0 (changes requested):
changes required

lsh-0:
I'm also not sure what marking anything as 'experimental' gets us if we're also using 'deprecated' and we only support two versions. Are we saying the spec may change? 

'experimental' would make more sense if we supported three versions - past (deprecated), current (stable), future (experimental).

I don't have a problem with that per se, but as a blanket policy it might not be suitable for all content types.

nlisgo:
> 'experimental' would make more sense if we supported three versions - past (deprecated), current (stable), future (experimental).

I intend to followup with the removal of `experimental` once I'm convinced the downstream changes have what they need from the new change. The `experimental` means I can make breaking changes if I've missed something. In the followup, I will add the `deprecated` flag to the version 3.

src/samples/press-package/v4/complete.json
nlisgo: > If this is a `complete.json` file, would it be good to include a `versionDate` here too?

Despite the `complete.json` label it's rarely complete where related content is concerned. I, at least, try to ensure one of every top level property declared in the model is available. I normally, when introducing a property, just start from the previous version samples and add the new thing. I've added `versionDate` anyway here.

nlisgo:
I'm going to reintroduce version 2. Have 3 versions while one of the versions is experimental. Mark version 2 as deprecated, although it's going to be a really small window.

When I followup with the removal of the experimental flag, I will remove version 2 and mark version 3 as deprecated.

lsh-0:
Sounds good.

PR #302, opened 2023-06-26
Add reviewed-preprint to /related endpoint

smoqadam:
Related:

* https://github.com/elifesciences/issues/issues/8411
* https://github.com/elifesciences/issues/issues/8373

- [x] update changelog


src/samples/article-related/v2/complete.json
```diff
@@ -0,0 +1,106 @@
...
...ts the E2 response appropriately\\nin the relevant cell types (endothelial cells and macrophage cells). Then they provide evidence for the MdM2-mediated post-translational regulation of ERa as a mechanism by which iron status impacts differential E2 has differential effects on ERa. Finally, they test the impact of systemic iron chelation on the OVX ApoE KO mice model of atherosclerosis and show that iron chelation attenuates the deleterious effects of E2 in late postmenopausal mice.\\n\\nOverall the evidence is solid and logically laid out. Given that serum iron levels do not correlate with the rest of the story, inclusion of LysM-macrophage-specific Fpn1 KO provides the key evidence that iron loading in macrophages drives differential effects to E2 in postmenopausal mice. The paper provides evidence that iron levels influence how the relevant cells respond to E2 with clinical implications to hormonal replacement therapy in younger and older postmenopausal women. While this study is limited by a small clinical sample size, it provides an important framework for future studies on the impact of age-related iron status on the response to hormonal replacement therapy.\\n\\nReferences",
```
nlisgo: Get rid of `indexContent` property and add `"type": "reviewed-preprint"`

src/model/article-related.v2.yaml
```diff
@@ -0,0 +1,16 @@
+$schema: http://json-schema.org/draft-04/schema#
+title: Related articles
+type: array
+items:
+    oneOf:
+      - $ref: ../snippets/article-poa.v1.yaml
+      - $ref: ../snippets/article-vor.v1.yaml
+      - $ref: ../snippets/reviewed-preprint.v1.yaml
```
nlisgo: reviewed-preprint items will need the `"type": "reviewed-preprint"` property.

lsh-0: should be good now

src/samples/article-related/v2/complete.json
```diff
@@ -0,0 +1,106 @@
...
+        "reviewedDate": "2022-10-20T03:00:00Z",
+        "versionDate": "2022-10-20T03:00:00Z",
+        "statusDate": "2022-10-20T03:00:00Z",
+        "subjects": [
+            {
+                "id": "cell-biology",
+                "name": "Cell Biology"
+            }
+        ]
+    }
```
lsh-0, line 105: example is missing some optional fields (titlePrefix, volume, eLocationId, curationLabels, image) but I don't think those are showstoppers.

src/model/article-related.v2.yaml
```diff
@@ -0,0 +1,22 @@
+$schema: http://json-schema.org/draft-04/schema#
+title: Related articles
+type: array
+items:
+    oneOf:
+      - $ref: ../snippets/article-poa.v1.yaml
+      - $ref: ../snippets/article-vor.v1.yaml
+      - allOf:
+          - properties:
+                type:
```
lsh-0, line 10: I guess I'd like to see every bit of content we serve up to self-identify with both `id` and `type` properties (and maybe even a `type-version`!), even if it's perfectly obvious from the context it was requested in. That won't be happening here so this shim will have to do.

lsh-0 (approved):

gnott (approved):
The data model looks good to me, a fine job.

I'm not too familiar with the `/reviewed-preprints` endpoint, and looking at it only briefly now, my comment it to be careful about how the articles are linked together, with respect to the `doi` values. If the `doi` is a non-eLife DOI -- a preprint server DOI -- when an eLife DOI already exists, then would the related article link sent a user to a non-eLife site when the click on it?

E.g. https://api.elifesciences.org/reviewed-preprints/90290  at this time has `"doi":"10.1101/2023.03.30.534478"`, but there is currently an eLife page https://elifesciences.org/reviewed-preprints/90290 and an eLife DOI (10.7554/eLife.90290, or 10.7554/eLife.90290.1) exists, not yet found in the API endpoint data.

lsh-0:
Yes, I'm a bit fuzzy on all that too, @gnott .

My assumptions are:

1. a reviewed preprint `doi` could point anywhere
2. a reviewed preprint `id` is also a manuscript ID and may (will?) become the `90290` part of a `10.7554/eLife.90290` doi.
3. the reviewed preprint `id` / manuscript ID may not exist yet in `lax`. It may not ever exist in lax.
4. the elife DOI generated from a reviewed preprint `id` may not exist yet either.

I'm not sure what the production process is for minting elife DOIs for preprints but the only thing I'm relying on is point 2: that a reviewed preprint ID is also a manuscript ID.

I've asked the question here https://github.com/elifesciences/issues/issues/8411#issuecomment-1670731074-permalink about detecting from the XML whether a reviewed preprint exists from a `related-article` element that is using a **doi**, but I *suspect* I can't trust that doi to exist, will have to extract it's manuscript ID and use that to query the `/reviewed-preprints` endpoint.

But then we get into muddy waters - do we still fetch the reviewed preprint if the related article is now published? Do we display both the reviewed preprint snippet and vor snippet in the `/articles/$id/related` endpoint? When both the reviewed-preprint content and content type changes, how do we keep dependent content updated? (does EPP push content updates to the message bus that we can listen for?)

This gets into the larger question of how EPP is being integrated into continuum and how tightly/loosely that should be. I've heard Paul wants a looser coupling because EPP is destined for more generic things, but if we try to accommodate that as well as accommodate this and the dozen other minor points of integration then EPP will essentially be part of the 'continuum' set of microservices, just a poorly integrated one. I'm tempted to write a 'dummy EPP' app just so we can get some end2end tests happening.

gnott:
I forgot the reasons too, but I think at least one reason is issue https://github.com/elifesciences/issues/issues/8411, to link Insight articles to EPP articles, perhaps.

lsh-0:
Yes, that's the ticket/epic I'm working on right now that is responsible for this PR. That epic was forked from [this epic](https://github.com/elifesciences/issues/issues/8373) for some reason, which has a bit more detail. The ostensible reason is to get related articles appearing under a header in the sidebar in the journal, which is already possible if external article markup is used in the XML instead.

gnott:
Probably the safest hack would be to concatenate the DOI using the manuscript ID, then it will always be correct, and ignore the DOI value from the API endpoint, but only because of the changable data there is I see right now.

lsh-0:
> Probably the safest hack would be to concatenate the DOI using the manuscript ID, then it will always be correct, and ignore the DOI value from the API endpoint,

Just so I'm clear, are you saying fetch the reviewed preprint and then replace the `doi` value we find there with an elife doi we build using the `id` value?

Would this help an elife-bot process somehow? I'm inclined to leave the `doi` as it is (all of that data as-is actually).

**edit**: 

> (all of that data as-is actually).

err, except `indexContent`, we're definitely dropping that ;)

gnott:
Sorry I've confused the whole thing. I'll try again though.

For example, insight XML https://github.com/elifesciences/elife-article-xml/blob/master/articles/elife-90447-v1.xml, has

```xml
<related-article related-article-type="commentary-article" ext-link-type="doi" xlink:href="10.7554/elife.80090" id="ra1"/>
```

That would be normal, except if the ~article  is still an RPP~ article it links to is still an RPP/RP/EPP status, and not yet a VOR, then you'd probably need to get metadata about it from the reviewed preprints API endpoint. It may not be listed under DOI `10.7554/elife.80090`, so instead it will probably be found by searching manuscript ID `80090`.

When the VOR comes along, I guess the related article link in Lax will be updated to use its metadata?

Edit: bad example maybe though, I don't know which Insight article has linked to an RP/EPP article yet.

lsh-0:
Don't worry about it, this is a good place to thrash things out.

> if the article ~is still an RPP article~ it links to is still an RPP/RP/EPP status, and not yet a VOR, then you'd probably need to get metadata about it from the reviewed preprints API endpoint. It may not be listed under DOI 10.7554/elife.80090, so instead it will probably be found by searching manuscript ID 80090.

My plan of attack will be to assume that *every* article from a certain date forwards may have had a reviewed preprint related article and query the /reviewed-preprints endpoint for it. If /reviewed-preprints responds with a 404 I'll know it didn't. I'll then backfill all articles since this certain date and lax will have all it needs. This is extra overhead for bot-lax, which is why it would be nice if we could know from the xml if a related article has a reviewed-preprint or not.

> When the VOR comes along, I guess the related article link in Lax will be updated to use its metadata?

Lax will need to decide to show either the reviewed preprint, the vor snippet or both. If we go back to the original intent - to display a link in a sidebar to the reviewed preprint - then I'm guessing the reviewed preprint snippet would override the vor snippet?

I haven't even bent my head around how reverse related reviewed preprints will work yet..

lsh-0:
> If we go back to the original intent - to display a link in a sidebar to the reviewed preprint - then I'm guessing the reviewed preprint snippet would override the vor snippet?

This is actually the purview of `recommendations` - deciding which items from a collection of related content to display: https://github.com/elifesciences/recommendations/blob/0114e16f027332eac8f7ee0739b1fc6c1711ef46/src/bootstrap.php#L123-L280

And the first thing it does is fetch these related articles:
https://github.com/elifesciences/recommendations/blob/0114e16f027332eac8f7ee0739b1fc6c1711ef46/src/bootstrap.php#L214

In which case I might just have lax return everything it has and let recommendations make that call.

lsh-0:
bump. @nlisgo, can we get your feedback (again) please?

nlisgo:
Wow, this has progressed. I'm happy for this to get in. If it proves too tricky to implement we can fallback to using external-article.

nlisgo:
@lsh-0 will leave you to merge in.

lsh-0:
cheers, @nlisgo , merging

PR #303, opened 2023-07-17
Add optional versionDoi to article-vor

nlisgo:


fred-atherden (approved):

gnott (approved):
With the assumption only VOR articles will have a `versionDOI`.

nlisgo:
> With the assumption only VOR articles will have a `versionDOI`.

If this changes we can extend to POA but in practice this is only going to be presented in VOR that has come through the new model. (VOR PRC)

gnott:
Checking with Fred, it seems that POA articles will not have a version DOI. Thought maybe in future you'll want it added to `snippets/reviewed-preprint.v1.yaml`.

nlisgo:
> Thought maybe in future you'll want it added to `snippets/reviewed-preprint.v1.yaml`.

We may revisit the reviewed-preprints at some point but currently this is a value that we only need in articles that are rendered in journal. EPP client does not use api-raml.

src/samples/article-history/v2/prc.json
```diff
@@ -33,6 +33,7 @@
           "version": 1,
           "type": "research-article",
           "doi": "10.7554/eLife.1234567890",
+          "versionDoi": "10.7554/eLife.1234567890.4",
```
lsh-0: `version` here is `1` and `versionDoi` is `4` ... going back through the tickets it looks like the two attributes have no relationship and that `versionDoi` is the version of the preprint before it became a VOR ...?  I think?

I vaguely remember pointing out that these dois would be ambiguous with the sub-dois we once used for abstracts. I think *that* ship has sailed, but is it too late to change this `versionDoi` field to something more self evident?

* `preprintDoi`
* `preprintVersionDoi`
* `eppDoi`
* `eppVersionDoi`

?

lsh-0: `reviewedPreprintDoi` ?

lsh-0: I've gone back to the XML:
https://github.com/elifesciences/XML-mapping/blob/bc20d5db27598b8529288ef8a1573350a2a0ebee/elife-1234567890-v2/elife-1234567890-v2.xml#L21-L23

And to what JATS have to say about `article-id` and `specific-use`:
* https://jats.nlm.nih.gov/publishing/tag-library/1.1d2/element/article-id.html
* https://jats.nlm.nih.gov/publishing/tag-library/1.1d2/attribute/specific-use.html

And I think `doiVersion` would actually be best here. There wouldn't be any association with epp/preprints and by putting the subject (`doi`) first it makes the `Version` portion more obvious it's qualifying `doi`, rather than qualifying `version` .

Does anybody have any objections to that? I'm happy to make the updates and apologise for being a nuisance.

nlisgo: > I've gone back to the XML: https://github.com/elifesciences/XML-mapping/blob/bc20d5db27598b8529288ef8a1573350a2a0ebee/elife-1234567890-v2/elife-1234567890-v2.xml#L21-L23
> 
> And to what JATS have to say about `article-id` and `specific-use`:
> 
> * https://jats.nlm.nih.gov/publishing/tag-library/1.1d2/element/article-id.html
> * https://jats.nlm.nih.gov/publishing/tag-library/1.1d2/attribute/specific-use.html
> 
> And I think `doiVersion` would actually be best here. There wouldn't be any association with epp/preprints and by putting the subject (`doi`) first it makes the `Version` portion more obvious it's qualifying `doi`, rather than qualifying `version` .
> 
> Does anybody have any objections to that? I'm happy to make the updates and apologise for being a nuisance.

Happy to switch to `versionDoi` in the schema. I will do the refactor, if I get a chance, today. We will adjust in our api-dummy PR.

nlisgo:
@smoqadam @viktoria-mahmud I've done the followup work on the api-dummy PR to change `versionDoi` to `doiVersion`.

https://github.com/elifesciences/api-dummy/pull/242

@lsh-0 I will merge this in in the morning so that @smoqadam is unblocked to progress further on prepping journal for this change.

lsh-0:
Thanks @nlisgo , I appreciate it.

PR #308, opened 2023-08-29
/articles, changes 'id' type from 'string' to 'int'

lsh-0:
- [x] update CHANGELOG


lsh-0:
Fred reminded me in an email the other day that we are approaching the switch to 6 digit MSIDs and that zero padding won't be used going forwards.

I've never supported string msids in lax and an msid of `00003` and `3` are treated the same. This change just reinforces that.

This *shouldn't* cause any problems for anyone, but I'm still a touch uncertain.

gnott (approved):
This will not cause any problems in the `elife-bot` usage, as long as the endpoint in the zero-filled style will still work, e.g. https://api.elifesciences.org/articles/00003

lsh-0:
yup, absolutely zero changes in current behaviour. This also works: https://api.elifesciences.org/articles/3

gnott:
I think `bot-lax-adaptor` may still be spitting out `id` as a string, which is returned by https://github.com/elifesciences/elife-tools/blob/dc49ed68f01be3a8617ee0b6704c392d9276ae0b/elifetools/parseJATS.py#L96-L98

lsh-0:
Internal code representations of the article ID can be anything. You could pass them around as strings, octals, hex, binary bytes etc, so long as when whatever app hits `https://api.elifesciences.org/articles/{id}` the `{id}` value is a decimal integer. 

Mathematically I don't know if `00003` and `3` are considered equal representations, but the current implementation of the API /articles endpoint and all of the msids in all of the references to articles at elife *ever* have never treated zero padding as significant.

PR #309, opened 2023-10-02
support for reviewed-preprint in collections

smoqadam:


src/api.raml
```diff
@@ -795,6 +795,15 @@ traits:
                                 complete:
                                     displayName: Complete
                                     value: !include samples/collection/v2/complete.json
+                        application/vnd.elife.collection+json;version=3:
```
nlisgo, line 788: this should be experimental and we could drop support for v1 at this stage and remove the experimental form v2.

nlisgo (changes requested):
Looks good but needs a little more effort.

PR #314, opened 2023-10-20
Add reviewed-preprint to recommendation

smoqadam:


src/api.raml
```diff
@@ -1789,6 +1789,20 @@ traits:
                                     Demonstration of the list being empty (the first page will always return a valid result).
                                 value: !include samples/recommendations/v2/empty.json
 
+                    application/vnd.elife.recommendations+json;version=3:
```
nlisgo, line 1778: Make experimental and we can delete version 1 as part of this PR.

nlisgo (changes requested):
We also need an entry in CHANGELOG.md

PR #316, opened 2024-02-14
Changes to funding data, adding awardDoi, optional recipients.

gnott:
Re issue https://github.com/elifesciences/issues/issues/8612

First draft includes adding `awardDoi` (https://github.com/elifesciences/issues/issues/8590), and removing `recipients` as being required (https://github.com/elifesciences/issues/issues/6876), to a new version of funding `funding.v2.yaml`.

That [new version of funding] is added to a next version `article.v4,yaml`.

That [new article.v4.yaml file] is used in new versions `article-poa.v4.yaml` and `article-vor.v8.yaml`.

poa v2 is removed, poa v3 is deprecated. vor v6 is removed, vor v7 is deprecated.

Duplicated the samples and an `awardDoi` example is included in `complete.json`,

```
{
    "id": "par-2",
    "source": {
        "funderId": "10.13039/100010269",
        "name": [
            "Wellcome Trust"
        ]
    },
    "awardDoi": "10.35802/212242"
},
```

Comments or changes welcome, @discodavey @lsh-0 @nlisgo. Plus if you think we can make these changes without having to version bump, that's fine too. I may have missed something here, and please correct me or leave some comments. Thanks!


lsh-0:
I've tweaked the PR description a little, @gnott , hope you don't mind.

> Plus if you think we can make these changes without having to version bump, that's fine too.

It's tricky, isn't it ;) I think all changes require version bumps almost all of the time, but the overhead is considerable so we fudge it a little bit to be practical. Like rolling these two features in together is practical. It makes things more complex but we should keep an eye out to see if we can do it again in future.

At a glance it seems fine. git-moving old examples to new examples (for example, article-vor v6 -> article-vor v8) may make it's git-history a bit schizo but this isn't code and it shouldn't come back to bite us.

I'll scrutinise it a bit more next week if that's ok? Adding myself as a reviewer.

lsh-0:
one thing, please update the CHANGELOG

gnott:
To mention as well, I opted for `awardDoi` instead of adding an `awardType` field, since it's probably good to use the DOI rules to validate the data.

It is very likely there will only be either `awardId` or `awardDoi`, never both, if we want to be restrictive in that way. I don't know the RAML that will enforce that. Plus, on the front-end the logic will still need to check for one or the other field values.

lsh-0:
> It is very likely there will only be either awardId or awardDoi

I've made a comment on the ticket https://github.com/elifesciences/issues/issues/8590 that I hope clarifies things a bit.

From my understanding of your solution there would be a new field added called `awardDOI` that would be a copy of `awardID` if `awardID` is a DOI. The `awardID` field wouldn't change to become optional if `awardDOI` were present.

Going back to the original description of that ticket, it sounds like they simply want to have links to funders where possible and it's not necessarily about DOIs (although with DOIs we can guarantee a working URL).

gnott:
The additional questions are great, keep it up @lsh-0! If the value is duplicated, then I should change the example in the sample JSON file `src/samples/article-vor/v8/complete.json`. However, if a URL is allowed, then there's more to change.

lsh-0:
Agreed. And if `awardID` becomes optional (assuming it's currently required) then there may be other effects we don't know about yet.

gnott:
> ... there may be other effects we don't know about yet.

Very true! I've updated the sample. It's easy to just add the new `awardDoi` value in the `elife-tools` parser and the `awardId` value remains unchanged.

lsh-0:
Fred has confirmed there are no plans for a non-DOI award URL in the future, so an `awardDOI` field should be fine.

CHANGELOG.md
```diff
@@ -1,5 +1,29 @@
 # Changelog
 
+## 2.28.0
+
+### Added
+
+* added `awardDoi` to new file `misc/funding.v2.yaml`, compared to `misc/funding.v1.yaml`
```
lsh-0: rm "compared to funding v1"

CHANGELOG.md
```diff
@@ -1,5 +1,29 @@
...
+* added `misc/article.v4.yaml` which uses `misc/funding.v2.yaml`
+* added `model/article-poa.v4.yaml` which uses `misc/article.v4.yaml`
+* added `model/article-vor.v8.yaml` which uses `misc/article.v4.yaml`
+* added support for `application/vnd.elife.article-vor+json;version=8`
+* added support for `application/vnd.elife.article-poa+json;version=4`
+
+### Changed
+
+* `recipients` no longer a required field in new file `misc/funding.v2.yaml`, compared to `misc/funding.v1.yaml`
```
lsh-0, line 16: awkward.

src/misc/article.v4.yaml
```diff
@@ -0,0 +1,87 @@
...
+        additionalFiles:
+            title: Additional files
+            type: array
+            items:
+                $ref: ../misc/asset-file.v2.yaml
+    required:
+      - copyright
+    dependencies:
+        authorLine: [authors]
+        authors: [authorLine]
```
lsh-0, line 87: diff

Screenshot at 2024-02-20 11-58-36

src/misc/funding.v2.yaml
```diff
@@ -0,0 +1,57 @@
...
+                              - name
+                    minItems: 1
+            required:
+              - id
+              - source
+    statement:
+        type: string
+        minLength: 1
+required:
+  - statement
```
lsh-0, line 57: diff 

Screenshot at 2024-02-20 11-59-26

src/model/article-vor.v8.yaml
```diff
@@ -117,10 +159,38 @@ allOf:
...
       - dependencies:
             authorResponse: [decisionLetter]
       - dependencies:
             authorResponse: [editorEvaluation]
+      - dependencies:
+            authorResponse: [elifeAssessment]
+      - dependencies:
+            authorResponse: [publicReviews]
+      - dependencies:
+            authorResponse: [recommendationsForAuthors]
```
lsh-0, line 196: diff

Screenshot at 2024-02-20 12-00-49

changes are misleading. suggest we do a regular copy of these in future rather than a git cp

src/samples/article-poa/v4/complete.json
```diff
@@ -137,6 +137,28 @@
         "Human",
         "Xenopus"
     ],
+    "image": {
```
lsh-0, line 140: diff

Screenshot at 2024-02-20 12-02-04

src/samples/article-vor/v8/complete.json
```diff
@@ -4,6 +4,7 @@
     "version": 1,
```
lsh-0, line 4: diff

Screenshot at 2024-02-20 12-03-21

lsh-0:
alright, once we get past the noise from using 'git cp' everything looks fine.

lsh-0 (approved):

lsh-0:
@nlisgo, I've added you as a reviewer too. If your role has changed and this isn't something you do any more let us know and we'll carry on without it.

src/misc/funding.v2.yaml
lsh-0: that's interesting actually, `awardID` isn't required. I had assumed it was.

gnott: I think since older XML would not have an award ID value it was always optional, and perhaps even now there could be ones without it, not sure.

gnott:
A big thumbs up for you checking things over @lsh-0 and the edits. I do a `git rm` then a `git add`, and those really don't show up nicely in the log. The only way to avoid it would be to add files in one PR, then remove them in a subsequent PR.

lsh-0:
> I do a git rm then a git add, and those really don't show up nicely in the log.

Well that's weird, Github is showing them as being renamed which is why the diffs that are present are a bit old and misleading:

Screenshot at 2024-02-21 11-19-02

Anyway, I'm not too concerned. This could be some clever thing Github does by extrapolating on the changeset. /shrug

discodavey:
@gnott As per our Slack chat. Nathan is trying to keep on EPP so if you need someone to review the work then Saeed or Luke are your people unless you are needing Nathan's advice that noone else can answer.

Does this make sense.

Happy with the naming convention btw

gnott:
This may be ready to merge, if you have any final input please @lsh-0

lsh-0:
Go for it

PR #318, opened 2024-06-28
Add specificUse field to data reference.

gnott:
Proposed changes to add `specificUse` to references of type `data`, its value can be either `analyzed` or `generated`.

Based on early discussion about how if we add a new optional field we do not need to bump the version numbers.

I promoted the `[unreleased]` notes in the changelog to be a proposed version `2.29.0` release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,16 +6,16 @@
 
 * added `specificUse` to references of type `data` in `misc/reference.v2.yaml`
 
-### Changed
-
-* changed `id` URI parameter in `/articles/{id}` from `string` to `integer`.
-    - zero padded integers are still integers and are fine.
-
 ## 2.30.0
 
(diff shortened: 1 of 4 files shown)


gnott:
Re issue https://github.com/elifesciences/issues/issues/8801

Re epic issue https://github.com/elifesciences/issues/issues/6888

fred-atherden (approved):

de-code (approved):
Looks superficially good to me. But I have to admit that I don't have any knowledge related to the kind of changes in this PR to properly review.

scottaubrey (approved):
Again, the patch looks OK to me, and seems to match the described change. I followed the README and it compiles and all, but not sure what other ramifications would be or how to surface them.

ali-tm-amin (approved):
Looks good to me

PR #322, opened 2024-11-06
Add support for elifeAssessment terms in article vor snippet

nlisgo:
We would amend samples also. And I will work on reviewed preprints piece too.


CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+## 2.31.0
+
+### Added
+
+* add optional strength and significance terms to elifeAssessment
+* add optional elifeAssessment to reviewed preprints snippet
+
+### Changed
+
+* introduce elifeAssessment to article-vor snippet
```
giorgiosironi: `introduce optional`?

src/snippets/reviewed-preprint.v1.yaml
```diff
@@ -83,6 +83,12 @@ properties:
                 - thumbnail
             - required:
                 - social
+    elifeAssessment:
+        type: object
+        $ref: ../misc/elife-assessment-terms.v1.yaml
+        required:
```
giorgiosironi: The set of required properties on an `elifeAssessment` is inconsistent depending on whether it's a snippet of an `article-vor` or a snippet of a `reviewed-preprint`. This can be confusing when we model these in PHP code, as many properties can be null

If the `elifeAssessment` properties are not present on the API responses yet, could we consider a stricter contract so that they appear in an atomic change?

giorgiosironi: e.g. `title` and `content` required, `significance` and `strength` optional for the time being

src/model/article-vor.v7.yaml
```diff
@@ -83,29 +83,6 @@ allOf:
                     format: uri
             required:
               - content
-        elifeAssessment:
```
giorgiosironi: verified `article-vor.v6.yaml` and lower do not exist anymore and don't need to be maintained

giorgiosironi: So should v7 be retired? https://prod--gateway.elifesciences.org/articles/100068 returns v8 in `Content-Type`

src/snippets/reviewed-preprint.v1.yaml
LinaKind: We have reviewed this and strive for a single elifeAssessment schema which should have all properties as optional, in order to implement it easily from the producers. From the point of view of journal this would be a single php class.

CHANGELOG.md
nlisgo: Changed the wording to: move optional elifeAssessment to article-vor snippet

src/model/article-vor.v7.yaml
nlisgo: We should discuss this sometime with @gnott present. This will not be addressed here.

nlisgo: The reason I am touching v7 in this instance is because we are bringing this property into the snippet and I assumed it would create a class. I may reinstate this and test it out.

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+## 2.32.0
+
+### Added
+
+* add optional strength and significance terms to elifeAssessment
+* add optional elifeAssessment to reviewed preprints snippet
```
giorgiosironi, line 8: might become mandatory, when implemented consistently

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+## 2.32.0
+
+### Added
+
+* add optional strength and significance terms to elifeAssessment
```
giorgiosironi, line 7: should become mandatory, when implemented consistently

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+* add optional elifeAssessment to reviewed preprints snippet
+
+### Changed
+
+* move optional elifeAssessment to article-vor snippet
```
giorgiosironi, line 12: will never be mandatory

